PR #1389, opened 2025-08-23
feat: Support opening policy in new tab for easier navigation

### LINKED ISSUE #1387: [Feature] - Policy table issues
Basically the ability to open the policy in a new tab, probably a menu that opens on right click or a button on the row to indicate open in new tab.

Dhanus3133:
## What does this PR do?

Support opening policy in new tab for easier navigation

- Fixes #1387

## Visual Demo (For contributors especially)
image

- [x] I have self-reviewed the code (A decent size PR without self-review might be rejected).
- [ ] I have updated the developer docs in /docs if this PR makes changes that would require a [documentation change](https://trycomp.ai/docs). If N/A, write N/A here and check the checkbox.
- [ ] I confirm automated tests are in place that prove my fix is effective or that my feature works.


Marfuen:
@Dhanus3133 The styling here doesn't look great, can we make sure it looks like this: https://ui.shadcn.com/docs/components/dropdown-menu, use that component or equivalent from the UI library

Marfuen:
@Dhanus3133 Actually I think this will be confusing for users now that I think about it and won't think about right clicking, let's just make the title specifically clickable with an underline so it opens in a new tab.

Dhanus3133:
> @Dhanus3133 The styling here doesn't look great, can we make sure it looks like this: https://ui.shadcn.com/docs/components/dropdown-menu, use that component or equivalent from the UI library

In the image, that was local browser’s dialog. In the previous implementation, since it only used React Router, a new tab wouldn’t open when the user tried **Ctrl + left click** or **right click → Open in new tab**. Instead, it just generated the route and navigated to the page directly in the same tab.

With my change, I used `Link`, so now opening in a new tab works as expected.

Dhanus3133:
> @Dhanus3133 Actually I think this will be confusing for users now that I think about it and won't think about right clicking, let's just make the title specifically clickable with an underline so it opens in a new tab.

Got it, just to confirm, should this behavior apply to all tables or only to this specific one? Since it’s a shared component, I wanted to make sure before changing it globally. So the idea is: when the user hovers over the title, it should appear underlined/clickable and always open in a new tab by default, while in other places it should continue opening in the same tab. Is that right?

Marfuen:
@Dhanus3133 Yeah I think we can make it standard, let's just do this for this table for now

Dhanus3133:
image

@Marfuen Updated it! Also added an icon when been hovered.

PR #2720, opened 2026-04-30
fix(integrations): source OAuth initiator userId from auth context

tofikwest:
## Summary

- **Production regression fix.** Every OAuth integration flow (GitHub, GCP, AWS, Rippling, Linear, Slack, etc.) is currently broken in prod, redirecting with `error=session_mismatch&error_description=OAuth flow can only be completed by the user who initiated it.`
- The frontend hook posts `userId: ''` with a comment claiming the API will fill it in from the auth context, but `oauth.controller.ts` reads `userId` straight from the request body — so every `IntegrationOAuthState` row is written with an empty string.
- This was harmless until #2712 (yesterday) added a defense-in-depth session check that compares `session.user.id` to `oauthState.userId`. Empty string never matches a real user ID → every flow fails.
- Resolve `userId` via the `@UserId()` decorator, drop the field from `StartOAuthDto`, and stop sending it from the client. The defense-in-depth check from #2712 is preserved and now actually works as intended.

## Why the spec didn't catch it

The new callback tests in #2712 mock `oauthState` directly with `userId: 'user_1'` rather than exercising `startOAuth → callback` end-to-end, so the empty-userId write path was never exposed.

## Scope

Strictly in scope. Verified:
- `IntegrationOAuthState.userId` is read in only two places (the new check at `oauth.controller.ts:475` and a warning log at `:289`) — both will continue working with the real userId.
- The repository (`oauth-state.repository.ts`) only queries by `state` (the random token), never by `userId`.
- No usages in `comp-private`, `apps/portal`, or `apps/framework-editor`.
- Only one caller of `/v1/integrations/oauth/start`: the frontend hook, which is updated here.

## Deploy-order safety

- **API ships first**: old frontends keep working — extra `userId: ''` in the body is ignored, decorator wins.
- **Frontend ships first**: the old API would crash on missing body field — but in practice this monorepo deploys API + app together.
- No data migration needed: in-flight OAuth states with empty userId expire after 10 minutes via `deleteExpired()`.

## Test plan

- [x] `npx jest src/integration-platform/controllers/oauth.controller.spec.ts` — 19/19 passing
- [x] `npx turbo run typecheck --filter=@trycompai/api` — no new errors introduced (177 pre-existing in unrelated files, identical count before/after)
- [x] `npx turbo run typecheck --filter=@trycompai/app` — no new errors introduced
- [ ] Manual smoke test in staging: connect GitHub via OAuth as the same user who initiated the flow → expect success
- [ ] Manual smoke test: confirm session check still rejects when initiating in one browser and completing in another (the security goal of #2712)

### DIFF
--- a/apps/api/src/integration-platform/controllers/oauth.controller.spec.ts
+++ b/apps/api/src/integration-platform/controllers/oauth.controller.spec.ts
@@ -176,9 +176,8 @@ describe('OAuthController', () => {
       mockedGetManifest.mockReturnValue(undefined as never);
 
       await expect(
-        controller.startOAuth('org_1', {
+        controller.startOAuth('org_1', 'user_1', {
           providerSlug: 'nonexistent',
-          userId: 'user_1',
         }),
       ).rejects.toThrow(HttpException);
     });
@@ -189,9 +188,8 @@ describe('OAuthController', () => {
       } as never);
 
       await expect(
-        controller.startOAuth('org_1', {
+        controller.startOAuth('org_1', 'user_1', {
           providerSlug: 'datadog',
-          userId: 'user_1',
         }),
       ).rejects.toThrow(HttpException);
     });
@@ -219,9 +217,8 @@ describe('OAuthController', () => {
       });
 
       await expect(
-        controller.startOAuth('org_1', {
+        controller.startOAuth('org_1', 'user_1', {
           providerSlug: 'github',
-          userId: 'user_1',
         }),
       ).rejects.toThrow(HttpException);
     });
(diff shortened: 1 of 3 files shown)


tofikwest:
Addressed @cubic-dev-ai review feedback in `210a4f7e1`.

**The concern:** The `@UserId()` decorator throws a plain `Error` for API-key callers, which NestJS turns into a generic 500 — bad UX and poor monitoring signal.

**Fix:** Added `SessionOnlyGuard` between `HybridAuthGuard` and `PermissionGuard` on `startOAuth`. Non-session auth is now rejected upfront with a clean **403 Forbidden** and a clear message, before `@UserId()` is ever evaluated.

```ts
@UseGuards(HybridAuthGuard, SessionOnlyGuard, PermissionGuard)
```

Note: this is purely a UX/error-reporting improvement — non-session callers could never complete the OAuth flow anyway because the callback requires a real session via `checkSessionMatchesState`. This just rejects them with a proper HTTP exception at the start endpoint instead of a generic 500.

Tests still passing 19/19.

PR #2903, opened 2026-05-21
fix: harden cloud remediation and chat

tofikwest:
## Summary
- route manual AWS remediations (RDS encryption/root access keys) to guided-only previews and block execute
- normalize AWS remediation plans by removing unsupported S3 ACL calls, backfilling security group IDs, and validating EC2 security-group params without rejecting SecurityGroupRuleIds-only revokes
- make Trigger remediation tasks surface non-JSON and empty API responses instead of raw JSON parse errors
- require explicit remediation execute success before single, batch, and per-finding retry flows report a finding as fixed
- skip IAM no-MFA findings for AWS IAM users without console access/login profiles
- disable stored OpenAI Responses item references for assistant chat to avoid missing reasoning item errors on follow-up messages

## Investigation notes
- Recent main commits already addressed empty current/proposed JSON for configure-only plans and AWS Config service-linked-role AWSServiceName normalization.
- Root access keys are detected from the IAM credential report root row; IAM users can be empty while root access keys still exist.
- Cubic reported valid response-handling gaps in the first PR revision; this branch now fails empty/malformed execute responses instead of treating them as fixed.
- A follow-up audit found the per-finding batch Retry button used an older server action with the same loose status check; that path now uses the same strict execute-response classifier.
- Owen's IAM MFA report was valid: programmatic-only IAM users should not be flagged for missing console MFA. The scanner now checks GetLoginProfile before evaluating a user for MFA.
- Latest Cubic review also found two valid gaps: manual previews could require credentials too early, and SecurityGroupRuleIds-only revokes were over-validated. Both are fixed with tests.

## Verification
- bunx jest src/cloud-security/remediation.service.spec.ts src/cloud-security/aws-command-executor.spec.ts src/cloud-security/plan-normalizer.spec.ts src/cloud-security/plan-normalizer-aws-edge-cases.spec.ts src/cloud-security/manual-remediation.spec.ts src/cloud-security/providers/aws/iam.adapter.spec.ts src/assistant-chat/openai-options.spec.ts --passWithNoTests
- bunx vitest run src/trigger/tasks/cloud-security/api-response.test.ts src/trigger/tasks/cloud-security/execute-result.test.ts
- git diff --check
- bunx prettier --check <touched files>

## Typecheck
Attempted focused API/app typechecks, but they are blocked by existing unrelated repo errors: Better Auth package path type mismatch plus pre-existing app/API test fixture and controller spec drift.

### DIFF
--- a/apps/api/src/assistant-chat/assistant-chat.controller.ts
+++ b/apps/api/src/assistant-chat/assistant-chat.controller.ts
@@ -39,6 +39,7 @@ import { AssistantChatService } from './assistant-chat.service';
 import { buildTools } from './assistant-chat-tools';
 import type { AssistantChatMessage } from './assistant-chat.types';
 import { RolesService } from '../roles/roles.service';
+import { ASSISTANT_OPENAI_PROVIDER_OPTIONS } from './openai-options';
 
 @ApiTags('Assistant Chat')
 @Controller({ path: 'assistant-chat', version: '1' })
@@ -129,6 +130,7 @@ Important:
         system: systemPrompt,
         messages: await convertToModelMessages(messages),
         tools,
+        providerOptions: ASSISTANT_OPENAI_PROVIDER_OPTIONS,
         stopWhen: stepCountIs(5),
       });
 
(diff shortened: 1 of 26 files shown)


tofikwest:
@cubic-dev-ai review it

tofikwest:
@cubic-dev-ai review it

tofikwest:
@cubic-dev-ai review it

PR #3007, opened 2026-06-03
fix(app): service card counts only added evidence tasks (matches detail)

tofikwest:
## Problem
The service card in a cloud integration's grid shows an evidence-task count = **total mapped templates**, while the service detail page (PR #3006) shows only tasks **added to the org**. So a card can read **"2 evidence tasks"** while the detail shows **"1"** — a mismatch that confuses users.

## Fix
Thread the org's task templates through `ProviderDetailView → ServicesGrid → ServiceCard` and count only mapped tasks that have a live task in the org, so the card matches the detail.
- The count badge **hides entirely** when none are added (existing `taskCount > 0` guard).
- Falls back to total mapped tasks if no templates are provided (defensive; the grid always passes them).

## Tests
`ServiceCard.test.tsx` (3 tests): added-only count (`1 evidence task` not `2`), fallback to total when `addedTemplateIds` absent, and the hidden-count case when none are added. All pass.

App typecheck: my files are clean (pre-existing `.test`/`auth-client` errors on `main` are unrelated, not in this diff).

## Related
- #3006 — hides "Not added" rows in the evidence lists (service detail + provider). This PR makes the **card count** consistent with that.
- #3003 — backend fix so AWS evidence checks actually run.

---
## Summary by cubic
Fixes the service card evidence-task count to show only tasks added to the org, matching the service detail page. Also fixes the fallback so cards show total mapped tasks when templates aren’t provided.

- Bug Fixes
  - Counts only mapped tasks with live org tasks via addedTemplateIds; hides the badge when zero.
  - Keeps addedTemplateIds undefined when taskTemplates is omitted so fallback to total works; an explicit empty array means “none added.”
  - Tests cover ServiceCard and ServicesGrid: added-only, fallback when omitted, and hidden badge.

<sup>Written for commit e47a500b04c34b47015c6930fdb4f46766ec5b25. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3007?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/app/src/app/(app)/[orgId]/integrations/[slug]/components/ProviderDetailView.tsx
+++ b/apps/app/src/app/(app)/[orgId]/integrations/[slug]/components/ProviderDetailView.tsx
@@ -397,6 +397,7 @@ export function ProviderDetailView({
                     connectionId={selectedConnection?.id ?? null}
                     orgId={orgId}
                     slug={provider.id}
+                    taskTemplates={taskTemplates}
                   />
                 )}
               </div>
--- a/apps/app/src/app/(app)/[orgId]/integrations/[slug]/components/ServiceCard.test.tsx
+++ b/apps/app/src/app/(app)/[orgId]/integrations/[slug]/components/ServiceCard.test.tsx
@@ -1,0 +1,54 @@
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+
+vi.mock('@/hooks/use-integration-platform', () => ({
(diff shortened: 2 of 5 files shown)


tofikwest:
@cubic-dev-ai review it

PR #3060, opened 2026-06-08
feat(integration-platform): add Azure MySQL + PostgreSQL Flexible Server TLS checks

tofikwest:
## Customer report
"Azure SQL Database — TLS 1.2 enforced" shows **0 passed** while "Azure Storage" shows a count. Customer: @Yahya El Iraki (MYIOT SOLUTIONS), who runs **MySQL Flexible Server**.

## The ticket's suspected root cause is wrong
It blamed a "missing `ctx.pass()` summary" in the SQL check. It isn't — `sqlTlsCheck` (`azure/checks/sql.ts:57-63`) emits `ctx.pass()` per compliant server, identical to Storage (`storage.ts:75-85`).

## Actual root cause: a coverage gap
Both checks share `list → if length===0 return → per-resource pass()/fail()`. Storage found 2 accounts → "2 passed". The SQL check queries **`Microsoft.Sql/servers`** (Azure SQL Database); the customer has none — they run **`Microsoft.DBforMySQL/flexibleServers`** — so it returns at `servers.length === 0` → **0 results → "0 passed"**. Green "Passed" run badges only mean the run completed without failures. Not a display bug — we simply didn't scan their databases.

An audit (Azure + GCP/AWS) found the **direct sibling**, **PostgreSQL Flexible Server**, has the identical gap. This PR fixes **both**.

## What this adds
Two checks, both mapped to the existing TLS/HTTPS evidence task, both reading TLS from **server parameters** (these resource types have no top-level `minimalTlsVersion` like Azure SQL):

| Check | Resource type | api-version | Params | Compliant |
|---|---|---|---|---|
| `azure-mysql-flexible-tls` | `Microsoft.DBforMySQL/flexibleServers` | `2023-12-30` | `require_secure_transport` + `tls_version` (comma-set) | secure transport ON **and** every enabled `tls_version` is 1.2+ |
| `azure-postgresql-flexible-tls` | `Microsoft.DBforPostgreSQL/flexibleServers` | `2024-08-01` | `require_secure_transport` + `ssl_min_protocol_version` (single) | secure transport ON **and** SSL floor 1.2+ (**unset = compliant** — PG only allows 1.2/1.3) |

Both verified against Microsoft Learn + the Azure REST spec; Reader role suffices. Follow existing conventions: no-op on zero resources, explicit "could not verify" on a config-read failure (no false pass), evidence + remediation. Registered in the manifest (`checks[]` + service entries). `runAllChecks` runs all manifest checks (no service gate).

## Tests
`azure-checks.test.ts` (bun) — pure helpers + full checks via the mock-ctx harness for both engines: compliant, comma-set / single-value, secure-transport OFF, sub-1.2 floor, PG unset-floor = pass, config-read failure → "could not verify", zero servers → no-op. **All 41 Azure tests pass; package builds + typechecks clean.**

## Honest caveats
- **Not live-tested** against real Azure MySQL/PostgreSQL servers — proof is an impersonated run post-deploy. The PG `2024-08-01` api-version is doc-verified (ARM template ref) but not live-pinned; confirm on first live run (same as MySQL's `2023-12-30`).
- **Deferred** (real but larger / separate, flagged by the audit — not in this PR): Cosmos DB, App Service, Redis TLS; the `no-public-access` flexible-server siblings; GCP AlloyDB; and several Azure services marked `implemented:true` with **no check at all** (cosmos-db, app-service, virtual-machine, aks, container-registry, **defender** [enabledByDefault!], policy) — these show "phantom coverage" and need either a check or `implemented:false`. Worth a follow-up ticket.


tofikwest:
Thanks — valid catch by cubic. Fixed in bec693a33: `ssl_min_protocol_version` reads now go through `readConfig()` which returns a discriminated `{ ok, value }`. A thrown/failed read → `{ ok:false }` → explicit "Could not verify" finding; a successful read with an unset value → `""` → a legitimate TLS 1.2 floor (PostgreSQL Flexible Server denies <1.2 regardless) → compliant. The two are no longer conflated, and there's a regression test for the ssl-read-failure path.

PR #3133, opened 2026-06-12
fix(cloud-security): run AWS integration checks on our server (scheduled false-fails)

tofikwest:
## Problem

AWS integration checks (S3 Block Public Access, encryption, EC2, RDS, KMS, CloudTrail, IAM) **false-fail on the scheduled / auto-run paths but pass on the manual "Run"** — repeatedly (CS reports over the last ~2 weeks). Customer-visible error:

```
AccessDenied: ... is not authorized to perform: s3:GetBucketPublicAccessBlock ...
because no VPC endpoint policy allows the s3:GetBucketPublicAccessBlock action
```

## Root cause

It is **not a code bug** — it's the same check code with the same credentials. The only difference is **where it executes**, and therefore which network the AWS API call exits through:

| | Executes in | S3 call exits via | Result |
|---|---|---|---|
| Manual "Run" | **our server (ECS)**, in `comp-production-vpc` | our S3 endpoint (allows `s3:Get*`) | ✅ |
| Scheduled / auto-run | **Trigger.dev's runtime** (their VPC) | Trigger.dev's S3 endpoint (blocks cross-account) | ❌ |

AWS evaluates allow/deny based on the VPC the request exits from — which, for the scheduled path, lives in **Trigger.dev's** AWS account and is not ours to change. (Confirmed Trigger.dev is Cloud, not self-hosted: our infra only whitelists Trigger.dev's static IPs; there's no Trigger VPC in our account. Their PrivateLink/static-IP features don't cover public-S3 egress.)

This is AWS-specific: GCP/Azure/dynamic/legacy checks make plain HTTPS calls to public APIs with no VPC endpoint involved, so they work identically in either environment.

## Fix (AWS only)

When `providerSlug === 'aws'`, the scheduled task (`run-task-integration-checks`) and the auto-run task (`run-connection-checks`) **delegate execution to our server** via a new internal, service-token-only endpoint, then persist the returned result with the **existing shared logic**. Trigger.dev still owns scheduling, retries, and the long-run tolerance — only *where the AWS API calls run* changed.

- **New** `ConnectionCheckRunnerService.runChecks()` — runs a connection's checks on our server and returns the raw result (no persistence).
- **New** `POST /v1/integrations/internal/run-connection-checks/:connectionId` (`HybridAuthGuard + ServiceTokenOnlyGuard + PermissionGuard`) — excluded from public docs/MCP.
- **New** `run-checks-on-server.ts` — the thin fetch helper the AWS branch calls.
[...]

PR #3135, opened 2026-06-15
fix(cloud-security): honor finding exceptions in integration task checks

tofikwest:
## Problem

A customer (PRIMER) marked a deliberately-public S3 bucket as an **exception** (`primer-production-reports-bucket`, reason "reports deliberately public"), but the **"S3 — public access blocked" task check still failed** — both the "1 failed" card and the compliance task status — on every run after the exception was created.

Root cause: `FindingException` was honored **only** in the Cloud Tests findings view (`cloud-security-query.getFindings`). The integration **task-check** path (run paths + display) never consulted exceptions, so an excepted finding was hidden in one surface but kept failing the task in another.

## Fix — one source of truth, applied everywhere a finding becomes pass/fail

This is the scalable part: the exception key format + active-filter now live in **one** place, so a future change can't honor an exception in one surface and miss another.

- **New `ActiveExceptionSet` / `loadActiveExceptionSet`** (`cloud-security/finding-exceptions.ts`) — the single definition of "is this finding excepted?" (keyed by `connectionId :: checkId :: resourceId`, active = not revoked / not expired).
- **Cloud Tests refactored** to use it (removes the duplicate private lookup — the two systems are now matched by construction).
- **Manual run-check** + **scheduled Trigger task**: task status excludes excepted findings via shared `countEffectiveFailures` + `decideTaskStatus`, so a task whose only failures are excepted goes **done** — identical logic in both paths.
- **Task-check display** (`getTaskCheckRuns`): excepted results are flagged (`excepted: true`) and dropped from the run's `failedCount`/status; the UI renders them as a muted **"Exception"** row instead of a red issue.

## Safety

- **Additive** — with no active exceptions the behavior is byte-for-byte identical (existing tests pass unchanged).
- **Fail-safe** — if the exception lookup errors, it suppresses **nothing** (never hides a real finding).
- **Raw data untouched** — `IntegrationCheckResult.passed` and the persisted run counts stay raw (audit intact); exceptions only affect *derived* status + display.
- Consistency note: the scheduled path's old "execution error ⇒ task failed" was dropped so it matches the manual path — an execution error is indeterminate (it gates `integrationLastRunAt` → retry next tick), not a compliance failure.

## Verification

- ✅ 44 tests pass across 10 suites — incl. new specs for the shared helpers and exception scenarios (manual run goes **done** when the only finding is excepted; `getTaskCheckRuns` flags excepted + drops it from the failed count).
- ✅ Changed API + app files type-clean.
- The PRIMER case: with the existing active exception, the scheduled and manual runs will now mark the task **done** and the card will stop showing the bucket as failed.

## Notes

- Independent of (and rebased on top of) the now-merged AWS-on-server work (#3133); this is a distinct, broader fix — it applies to **all** providers' task checks, not just AWS.
- The UI addition reuses the file's existing `@trycompai/ui` `Badge` (didn't migrate the whole legacy file — out of scope).

---
## Summary by cubic
Fixes task checks to honor finding exceptions across manual runs, scheduled runs, and the UI. Excepted findings no longer fail tasks, and status/display are consistent while keeping real execution errors visible.

- **Bug Fixes**
  - Task status excludes excepted findings; a task with only excepted failures is marked done (even with 0 passing).
  - Run history flags excepted results, removes them from failed counts, and shows a muted “Exception”; failed → success rewrite happens only when failures were excepted, so execution-error runs stay failed.
  - Manual and scheduled paths use the same evaluation; lookup failures suppress nothing, and execution errors don’t drive status.
  - Added tests, including provider-agnostic coverage (AWS, GCP, Azure).

- **Refactors**
  - Introduced `ActiveExceptionSet`/`loadActiveExceptionSet` as the single source of truth; Cloud Tests now use it too.
  - Added shared `countEffectiveFailures` and `decideTaskStatus` used by both run paths.
  - Raw check rows and persisted counts are unchanged; only derived status and display are adjusted.

<sup>Written for commit bd1b774b48cde7cdeea856bc7421e99f40ae5f76. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3135?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/cloud-security/cloud-security-query.service.ts
+++ b/apps/api/src/cloud-security/cloud-security-query.service.ts
@@ -4,6 +4,7 @@ import { getManifest } from '@trycompai/integration-platform';
 import { sanitizeEvidence } from './evidence-sanitizer';
 import { getLegacyFindings } from './cloud-security-query.legacy';
 import { resolveCheckKey } from './check-definition.utils';
+import { loadActiveExceptionSet } from './finding-exceptions';
 import type {
   CloudFinding,
   CloudProvider,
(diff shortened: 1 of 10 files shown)


tofikwest:
@cubic-dev-ai review it

PR #3143, opened 2026-06-15
fix(integration-platform): serialize OAuth token refresh per connection

tofikwest:
## Problem

Customers' **Google Workspace** and **GCP** integrations show **"OAuth token expired. Please reconnect the integration."** every few days. Reported for one org, but it's **platform-wide**.

Verified directly against production (`compdb_production`, 2026-06-15):

| Provider | Active | In error | Distinct orgs in error |
|---|---:|---:|---:|
| GCP | 13 | **52** | 52 |
| Google Workspace | 161 | **158** | 158 |
| AWS (no OAuth refresh) | 256 | **0** | — |

AWS uses IAM roles (no OAuth refresh) and has **zero** errors — isolating the fault to our **OAuth token-refresh mechanism** for Google. The reported org has **no BYO OAuth app**; it uses our shared platform OAuth app, so this is our code, not a customer Google config.

## Root cause

`refreshOAuthTokens` has **no concurrency control**. The same connection gets refreshed by multiple workers at once:
- `run-integration-checks-schedule.ts` fans out **one Trigger job per task** mapped to a check on the **same** connection, each calling `/ensure-valid-credentials` → `refreshOAuthTokens`.
- `sync.controller.ts` **force-refreshes unconditionally** on every sync.

Concurrent refreshes of a single Google refresh token cause Google to reject it (`invalid_grant`), which is the only thing that writes the "OAuth token expired" status (`credential-vault.service.ts`, HTTP 400/401/403).

**Production signature confirming the race:** a successful refresh at ~06:00 UTC (the `0 6 * * *` scheduler) followed by an error ~4h later the same day (median 0 days from last success), and **128 of 314 Workspace connections show two refreshes within 60 seconds** (108 within 5s).

Ruled out: storage loss (carry-forward + keep-5 are correct), 7-day "Testing mode" expiry (healthy multi-month refresh chains exist), and the 100-token cap.

## Fix — per-connection distributed single-flight

Only one worker refreshes a given connection at a time; others wait and reuse its fresh token.

- **Schema/migration:** nullable `IntegrationConnection.refreshLeaseUntil` (additive, instant — no table rewrite).
- **`ConnectionRepository.acquireRefreshLease` / `releaseRefreshLease`:** atomic compare-and-swap (`UPDATE … WHERE lease IS NULL OR expired`).
- **`refreshOAuthTokens`:** wraps the existing logic (renamed `performTokenRefresh`) — acquire lease → if busy, poll & reuse the peer's token → else refresh → release in `finally`.

**Why a DB lease and not `pg_advisory_xact_lock`:** an advisory lock holds a DB connection for the entire token HTTP call; at the 06:00 thundering herd that risks pool exhaustion. The lease grabs/releases in two sub-ms UPDATEs, holds no connection during the HTTP, and auto-expires (180s) so a crashed worker can't wedge refreshes.

**Force-refresh preserved:** the 401-retry (`onTokenRefresh`) and `forceRefresh: true` paths still really refresh — coalescing only reuses a token produced **after** the call began. Lease errors degrade to a direct refresh.

## Testing

- New unit tests (3/3): reserved-param safety (existing), concurrent coalesce (no duplicate provider call), lease always released on failure.
- `@db` builds; `prisma validate` passes; changed files typecheck clean (repo baseline is pre-existing-RED on unrelated smithy/spec files).

## Not in this PR / follow-ups

- The exact Google error (`invalid_grant` reason / `error_subtype`) is only in app logs at `credential-vault.service.ts:448` — worth pulling for one errored connection to 100%-confirm the trigger.
- Add a fetch timeout to the refresh call (currently none → a hung refresh holds the lease for the full TTL).
- Consider gating the unconditional Workspace-sync refresh behind `needsRefresh`.
- An integration test against a real DB would prove the CAS lease end-to-end (unit tests cover the branching).

---
## Summary by cubic
Serialize OAuth token refresh per connection to stop concurrent refresh races that invalidated Google refresh tokens. Adds an ownership-checked DB lease, request timeouts, and safer fallbacks to remove false “OAuth token expired. Please reconnect.” errors across GCP and Google Workspace.

- **Bug Fixes**
  - Per-connection single-flight via DB lease (180s TTL); adds `refreshLeaseUntil` and fencing token `refreshLeaseToken`. Release is ownership-checked; no advisory locks held during HTTP.
  - 20s AbortController timeout on provider calls; network/timeout errors are transient (retry, don’t mark expired). Force-refresh/401 paths preserved.
  - Callers coalesce via polling and a post-acquire freshness check; only one provider call per connection.
  - If lease coordination fails (DB error), fail safe: return the stored token (no unserialized refresh).
  - Tests cover reserved-param safety, coalescing (pre/post-acquire), ownership-checked release, transient timeout, and lease-unavailable fail-safe.

- **Migration**
  - Add nullable `IntegrationConnection.refreshLeaseUntil` and `refreshLeaseToken` (additive, no rewrite; safe to deploy).

<sup>Written for commit dc4b3f4a5da88e298de6f23f2f836daf8caffd70. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3143?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tofikwest:
Good catch — fixed in `dc4b3f4a5`. You're right that `releaseRefreshLease` cleared the lease without an ownership check.

**Fix (fencing token):**
- `acquireRefreshLease(id, ttl, leaseToken)` now stamps a per-refresh `randomUUID()` owner token.
- `releaseRefreshLease(id, leaseToken)` only clears the lease `WHERE "refreshLeaseToken" = $token`, so a holder whose work outlived the TTL can no longer wipe a lease another worker has since acquired (the release just matches 0 rows).
- New `refreshLeaseToken` column (added to the same migration).

**Also hardened so that window can't realistically open:**
- 20s `AbortController` timeout on the token HTTP call → worst-case refresh ~42s (20s + 2s + 20s) stays well under the 180s lease TTL, so a hung provider call can't hold the lease long enough for another worker to re-acquire and overlap. Timeouts are treated as transient (retried, **not** marked "expired").
- On a lease-coordination failure (DB error), the path now **fails safe** — returns the stored token instead of doing an unserialized refresh (correlated DB failures would otherwise fan every fan-out caller into the provider at once, reintroducing the exact race).

**Validation:** added unit tests for the same-token acquire→release round-trip, post-acquire coalesce, lease-unavailable fail-safe, and timeout-as-transient (6/6 pass). The CAS mutual-exclusion + token-matched release live in raw SQL, so a real-DB integration test (two concurrent acquires → exactly one wins; wrong-token release is a no-op; expired lease re-acquirable) is the right place to assert the guarantee end-to-end — noted as a follow-up since there's no ephemeral-Postgres harness in the unit suite.

Separately flagged for a follow-up (pre-existing, not in this diff): `credentialRepository.create` assigns the version via a non-atomic read-then-write, so the OAuth connect/reconnect path can race a lease-holding refresh and hit `@@unique([connectionId, version])` (P2002). Best fixed at the create layer (P2002 retry or single-statement version assignment), not by extending the lease to the connect path.

tofikwest:
@cubic-dev-ai review it

PR #3144, opened 2026-06-15
fix(automation-builder): stop test dialog hanging on "Running your automation"

tofikwest:
## Problem

In the task automation **Integration Builder**, testing a script gets stuck on **"Running your automation"** forever and never returns a success/failure result. It reproduces even with a trivial no-op (`module.exports = async () => ({ ok: true })`), so the hang is in the result-delivery path, not the script.

## Root cause

The builder test delivers its result **only by 1-second polling** of the Trigger.dev run (`useTaskAutomationExecution` → enterprise-api `/runs/[runId]` → `runs.retrieve()`; there is no realtime/SSE channel). The poller only treated two statuses as terminal:

- `COMPLETED && output` → success
- `FAILED` → error
- **everything else** → `setTimeout(poll, 1000)` with **no ceiling**

So every *other* terminal Trigger.dev v4 run status — `TIMED_OUT`, `CRASHED`, `SYSTEM_FAILURE`, `EXPIRED`, `CANCELED` — plus `COMPLETED`-without-output and a run stuck in `QUEUED`, fell into the infinite re-poll. The most common path: a run hits the backend task's `maxDuration: 300` and ends `TIMED_OUT`, which the UI then polls forever.

> The matching backend hardening (time-boxing the AI calls that could push a run to `TIMED_OUT`) is in a companion PR on `comp-private` (same branch name).

## Fix (`use-task-automation-execution.ts`)

- Treat `COMPLETED` as success **regardless of output payload** (no more falling through on empty output)
- Surface a clear, status-specific error for **every** terminal-but-not-`COMPLETED` status; treat unknown/future statuses as terminal (fail-safe) rather than looping
- Only the genuinely in-progress statuses (`PENDING_VERSION/QUEUED/DEQUEUED/EXECUTING/WAITING/DELAYED`) re-poll
- Add an absolute **polling deadline backstop** (6 min, above the backend's 5-min `maxDuration`) so the dialog can never spin forever — even on a stuck-`QUEUED` run or a changed response shape

The live UI is `TestResultsPanel` (`TestDialog` is dead code); both read the same hook, so this single change covers the rendered panel.

## Tests

`use-task-automation-execution.test.tsx` — 11 cases: COMPLETED (with/without output), FAILED, each terminal-failure status (TIMED_OUT/CRASHED/SYSTEM_FAILURE/CANCELED/EXPIRED), unknown status, polls-through-in-progress-then-resolves, and the deadline backstop. All passing. `tsc --noEmit` clean for the changed file.

## Not fixed (pre-existing, out of scope — flagged)

- Summary call uses a bare `'gpt-4o-mini'` model id (canonical is `'openai/gpt-4o-mini'`) → likely 404s silently
- `shouldMarkTaskFailed` has no `version` gate, so a draft test of an *enabled* automation could mark its task `failed` if the AI evaluation returns 'fail'

## Verification note

Not reproduced live; the upstream trigger for the reporting org isn't prod-confirmed. The frontend fix resolves the symptom regardless of cause. To confirm which upstream cause hit `org_6a23d8b8a13d62f7001fa12c`, check the Trigger.dev runs of `execute-automation-script`: `TIMED_OUT` at ~300s ⇒ AI stall; never-advancing `QUEUED` ⇒ queue saturation.

---
## Summary by cubic
Fixes the Integration Builder test panel hanging on “Running your automation” by handling all terminal run statuses and adding an independent 6‑minute timeout. The dialog now always resolves with success or a clear error, even if a status request never returns.

- **Bug Fixes**
  - Treat `COMPLETED` as success even when `output` is missing; sanitize and surface script errors when present.
  - Stop polling on terminal statuses (`FAILED`, `TIMED_OUT`, `CRASHED`, `SYSTEM_FAILURE`, `CANCELED`, `EXPIRED`, unknown) with clear messages; only in‑progress statuses keep polling.
  - Add an independent 6‑minute backstop timer with a single‑resolution guard so a hung request or stuck queue can’t spin forever; add regression tests for terminal/unknown statuses, in‑progress flow, the timeout, and a hung status request.

<sup>Written for commit 771ae854bd3d3eee851b4aea626ead93e7966695. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3144?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tofikwest:
Companion backend PR (AI call time-boxing): trycompai/comp-private#293 — same branch `tofik/fix-automation-test-hang`.

PR #3165, opened 2026-06-16
fix(integration-platform): run dynamic integration checks on the daily schedule

tofikwest:
## Problem
Dynamic (DB-backed) integrations — **Keeper, Supabase, and ~575 others** — never ran on the daily schedule. They only ran when a user clicked **Run** manually, so their task "last run" sat stale for days. **1,167 active dynamic connections across 282 orgs** are affected (verified in prod).

## Root cause
The daily orchestrator (`integration-checks-schedule`) and worker (`run-task-integration-checks`) run in the **Trigger.dev runtime**, whose manifest registry is seeded with only the **8 static code manifests**. The dynamic-manifest loader is a NestJS `OnModuleInit` service that **never runs in Trigger.dev**, so `getManifest(dynamicSlug)` returns `undefined` and the orchestrator silently skipped those connections (`continue`). Manual runs work because they execute in the **API process**, where the loader has populated the registry.

Confirmed in prod on an example task: github/google-workspace (static) get the 06:00 UTC scheduled runs; Keeper ran once (manual), Supabase ran once (manual).

## Fix (mirrors the existing AWS server-delegation)
- **Discovery** — the orchestrator now reads dynamic task-mappings straight from the DB (`DynamicCheck.taskMapping` / `isEnabled`) via `resolveProviderChecks`. Static providers still resolve from the code manifest and win when both exist.
- **Execution** — the worker and auto-run-on-connect delegate dynamic providers to the API server (`runChecksOnServer` → the already provider-agnostic `ConnectionCheckRunnerService`), where the dynamic manifest is loaded. `shouldRunOnServer = aws || (no local manifest && active dynamic)`. A truly unknown provider still early-returns instead of crashing.

**Static integrations and AWS are unchanged** — the in-process path only runs when a manifest is present. Also fixes scheduled auto-run-on-connect for dynamic providers.

## Why it's safe (no regression to working behavior)
- Static + AWS paths are byte-identical no-ops (`runOnServer` only adds a branch).
- Manual-run path is untouched.
- Risk audit of every `api_key`/`basic` dynamic integration platform-wide: only `coda`/`pinecone` have ever run successfully — both provably safe (a successful run proves the credential resolves under the same lookup the server runner uses). Everything else has never run. `custom`/`oauth2` (incl. Keeper/Supabase) are byte-identical to the proven manual path.
- Load when enabled: ~2,775 server check-calls/day — trivial; no concurrency cap added (a task-level cap would have throttled AWS, which shares the worker).

## Tests
- New `dynamic-provider.spec.ts` (`shouldRunOnServer` + `isActiveDynamicProvider`) and `resolveProviderChecks` cases.
- `cd apps/api && npx jest src/trigger/integration-platform` → 5 suites / 21 tests pass.
- API typecheck: changed files clean (pre-existing `main`-RED errors unrelated: smithy dedup + framework-editor spec drift).

## Follow-up (not in this PR)
Scheduled employee sync (`sync-employees-schedule.ts:64`) has the identical `getManifest` gate for dynamic providers — separate fix.

---
## Summary by cubic
Fixes missed daily runs for dynamic (DB-backed) integrations by reading their checks from the DB and delegating execution to the API server when no local manifest exists. Static integrations and AWS are unchanged; manual runs are unaffected.

- **Bug Fixes**
  - Discovery: Orchestrator uses `resolveProviderChecks` to fall back to DB-backed `dynamicIntegration.checks` (enabled only) when `getManifest` is missing; static manifests win.
  - Routing: Worker and auto-run use `shouldRunOnServer` to delegate when `aws` or (no manifest + active dynamic) and call `runChecksOnServer`; unknown providers no-op.
  - Credentials/vars: Skip Trigger-side credential preflight only for delegated runs; in-process path keeps token refresh and required-variable checks.
  - Error handling: Server-run transport failures are recorded as per-check errors (with retry gating) instead of aborting the task; in-process behavior unchanged.
  - Tests/hygiene: Added `dynamic-provider.spec.ts` and `resolveProviderChecks` tests; unanchored `.gitignore` rules for `dbq_*.js` and `*.csv`.

<sup>Written for commit c4f56a19cdd250903cc1d012c4d19944b9d39de6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3165?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,6 +105,6 @@ scripts/sync-release-branch.sh
 apps/mcp-server/.github/
 
 # Ad-hoc read-only prod ops scripts (contain DB creds) + any exported customer data.
-# These must never be committed.
-/dbq_*.js
-/*.csv
+# These must never be committed. Unanchored so nested paths are covered too.
+dbq_*.js
+*.csv
--- a/.gitignore
+++ b/.gitignore
@@ -103,3 +103,8 @@ scripts/sync-release-branch.sh
 # from .github/workflows/ at the repo root. We move them up. See
 # .github/workflows/sdk_generation.yaml + sdk_publish.yaml (the real ones).
 apps/mcp-server/.github/
+
+# Ad-hoc read-only prod ops scripts (contain DB creds) + any exported customer data.
+# These must never be committed.
+/dbq_*.js
+/*.csv
--- a/apps/api/src/trigger/integration-platform/dynamic-provider.spec.ts
+++ b/apps/api/src/trigger/integration-platform/dynamic-provider.spec.ts
@@ -1,0 +1,62 @@
+// Mock @db so importing the helper doesn't open a Postgres connection.
+const findUnique = jest.fn();
+jest.mock('@db', () => ({
+  db: { dynamicIntegration: { findUnique: (...args: unknown[]) => findUnique(...args) } },
+}));
+
+import { isActiveDynamicProvider, shouldRunOnServer } from './dynamic-provider';
+
+describe('shouldRunOnServer', () => {
+  it('always delegates AWS (VPC egress), with or without a manifest', () => {
+    expect(
+      shouldRunOnServer({ providerSlug: 'aws', hasManifest: true, isActiveDynamic: false }),
+    ).toBe(true);
+    expect(
+      shouldRunOnServer({ providerSlug: 'aws', hasManifest: false, isActiveDynamic: false }),
+    ).toBe(true);
+  });
+
+  it('runs static providers (manifest present) in-process', () => {
+    expect(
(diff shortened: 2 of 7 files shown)


tofikwest:
@cubic-dev-ai review it

PR #3176, opened 2026-06-17
feat(integrations): cloud evidence-task checks — GCP M&A/Encryption + Separation of Environments (GCP/AWS/Azure)

tofikwest:
## Problem

A customer (org `org_6a0f40b752e63d6afc3c1a42`) with GCP connected reported that GCP doesn't appear on many evidence items where **AWS and Azure do** — e.g. the **Monitoring & Alerting** task showed AWS + Datadog but no GCP.

**Root cause (not pagination):** the task integration picker (`getChecksForTaskTemplate`) includes an integration's check only when `check.taskMapping === task.taskTemplateId`. GCP shipped just 5 checks covering 4 templates (`tlsHttps`, `backupLogs`, `productionFirewall`, `rolebasedAccessControls`) — it had **no** check for `monitoringAlerting`, `encryptionAtRest`, or `secureSecrets`, which AWS/Azure have. So GCP was structurally absent from those tasks. It's a coverage gap, not a bug in the picker.

## Change

Three new GCP checks, built on the existing GCP check runtime (`resolveGcpProjectIds`, `gcpListItems`, `ctx.fetch`, `isGcpApiDisabled`, `toHttpReadFailure`):

| Check | Task | Behavior |
|---|---|---|
| `gcp-cloud-monitoring-alerting` | Monitoring & Alerting | Two prongs per project (mirrors Azure `monitor.ts`): (1) an **enabled alert policy wired to a notification channel**; (2) a **configured log-export sink** beyond the managed `_Default`/`_Required`. An unreadable prong fails "could not verify" — never a silent pass. |
| `gcp-storage-encryption` | Encryption at Rest | GCP always encrypts at rest, so it **passes each bucket** and records the key type (Google-managed vs CMEK via `encryption.defaultKmsKeyName`). Only read failures emit a finding. |
| `gcp-cloud-sql-encryption` | Encryption at Rest | Same, for Cloud SQL instances (CMEK via `diskEncryptionConfiguration.kmsKeyName`). |

**GCP coverage: 4 → 6 task templates (5 → 8 checks).**

## Manual + scheduled both verified

Traced both execution paths in code — they run my checks through the **same engine** (`runAllChecks → runCheck → createCheckContext → check.run(ctx)`):
- `ctx.fetch` attaches the OAuth bearer token to absolute `*.googleapis.com` URLs (`buildUrl`/`buildHeaders`), with `401 → onTokenRefresh → retry`.
- Scheduled: `shouldRunOnServer('gcp')` is **false** (only AWS + dynamic delegate to the server), so GCP runs **in-process in the Trigger.dev runtime** and calls **public** Google APIs — unaffected by the AWS-VPC and dynamic-integration scheduler constraints. GCP is a static manifest, so the orchestrator's `getManifest('gcp')` resolves these checks.

## Tests
- **20 new tests** (both prongs, read-failure → "could not verify", API-disabled skip, `enabled` default, CMEK vs Google-managed evidence).
- Green: 332 package tests, 21 trigger/scheduler tests, 10 manual-path controller tests, typecheck, build.

## Notes for reviewer
- **Not yet live-tested against a real GCP project** — verification plan is mock tests (done) + customer live run. Behavior is identical whether run manually or by the cron.
- **IAM:** the Monitoring & Alerting check needs `monitoring.viewer` + `logging.viewer` (or `Viewer`) on the connection, else it returns a graceful "could not verify" with the exact grant remediation. The encryption checks need `storage.buckets.list` + `cloudsql.instances.list`, already used by the existing GCP checks.
- **"Separation of Environments"** (also raised by the customer) is intentionally out of scope — no check maps to it for *any* provider; that's a separate product decision on what signal proves env separation.

---
## Summary by cubic
Adds three GCP checks to surface GCP in the Monitoring & Alerting and Encryption at Rest tasks. GCP coverage expands from 4 to 6 task templates (5 → 8 checks).

- **New Features**
  - `gcp-cloud-monitoring-alerting` (→ `monitoringAlerting`): per project, requires an enabled alert policy wired to a notification channel and at least one non-default, enabled log export sink. If unreadable, reports “could not verify”; skips when APIs are disabled.
  - `gcp-storage-encryption` (→ `encryptionAtRest`): passes each bucket and records key type (Google-managed or CMEK). Only read failures emit a finding.
  - `gcp-cloud-sql-encryption` (→ `encryptionAtRest`): passes each instance and records key type (Google-managed or CMEK). Only read failures emit a finding.

<sup>Written for commit d294ba0312015dd3fa30abbb50f8644d5cec0d52. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3176?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/integration-platform/src/manifests/gcp/checks/__tests__/gcp-checks.test.ts
+++ b/packages/integration-platform/src/manifests/gcp/checks/__tests__/gcp-checks.test.ts
@@ -4,15 +4,22 @@ import type {
   CheckVariableValues,
   IntegrationCheck,
 } from '../../../../types';
+import { cloudMonitoringAlertingCheck } from '../cloud-monitoring-alerting';
 import { cloudSqlBackupsCheck } from '../cloud-sql-backups';
+import { cloudSqlEncryptionCheck } from '../cloud-sql-encryption';
 import { cloudSqlSslCheck } from '../cloud-sql-ssl';
 import { iamPrimitiveRolesCheck } from '../iam-primitive-roles';
+import { storageEncryptionCheck } from '../storage-encryption';
 import { storagePublicAccessCheck } from '../storage-public-access';
 import { vpcOpenFirewallsCheck } from '../vpc-open-firewalls';
 import { isGcpApiDisabled } from '../shared';
 
 interface Captured {
-  passed: Array<{ resourceId: string; title: string }>;
+  passed: Array<{
+    resourceId: string;
+    title: string;
+    evidence?: Record<string, unknown>;
+  }>;
   failed: Array<{
     resourceId: string;
     title: string;
@@ -43,7 +50,12 @@ async function runCheck(
     log: () => {},
     warn: () => {},
     error: () => {},
-    pass: (r) => passed.push({ resourceId: r.resourceId, title: r.title }),
+    pass: (r) =>
+      passed.push({
+        resourceId: r.resourceId,
+        title: r.title,
+        evidence: r.evidence,
+      }),
     fail: (r) =>
       failed.push({
         resourceId: r.resourceId,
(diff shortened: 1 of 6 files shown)


tofikwest:
### Update — scope expanded + review fix applied

Two changes since the initial review:

**1. Fixed the cubic finding (log-export prong).** The Monitoring & Alerting log-export prong now requires a **durable destination** (BigQuery / Cloud Storage / Pub/Sub, or a non-default Cloud Logging bucket) instead of just a non-default sink name — so a custom-named sink that still targets the `_Default` bucket no longer counts as export. Matches the Azure Monitor check's bar. +2 tests.

**2. Added a third GCP check: Separation of Environments** (`gcp-environment-separation` → `separationOfEnvironments`). This task previously had **no check for any provider**, so its picker was empty for everyone (the customer's "can't connect to any cloud provider" report).

⚠️ **Heuristic — please review the approach.** Unlike the others, "environments are separated" has no single API field. This check lists all accessible projects, classifies each into an environment (by `environment`/`env` label, else a word-boundary name/id token so `product`/`developer` don't false-match `prod`/`dev`), and **passes when ≥2 distinct environments are detected**. When it can't confirm (single env / unclassified / no projects) it **fails with actionable guidance** (label projects, or upload a diagram) rather than silently passing — the task still accepts manual evidence. It evaluates the whole project footprint, ignoring `project_ids` scoping.

**GCP coverage is now 7/7 automatable task templates** (9 checks); only Secure Secrets is left (deliberately deferred — weak signal). 346 package tests green, typecheck + build clean.

tofikwest:
### Update — Separation of Environments now covers AWS + Azure (parity)

Added the equivalent check for AWS and Azure (all three clouds now appear on the **Separation of Environments** task), plus a shared classifier and a review-driven bug fix.

**New `manifests/environment-classification.ts` (shared, like `http-read-failure`)** — token-exact classification: a name/tag is split on any non-alphanumeric separator (`-`, `_`, `.`, `/`) and tokens are matched *exactly*. This (a) keeps the precision (`product`≠`prod`, `developer`≠`dev`) and (b) **fixes a latent bug an adversarial review caught: the old `\b` regex treated `_` as a word char, so `myapp_prod` classified to `null`.** Reads only env-key tag/label values (never arbitrary tags) so a stray `team=dev-team` can't fabricate an environment. GCP's check was refactored onto it (so it gets the underscore fix too).

**AWS (`service: ec2-vpc`)** — classifies non-default, available VPCs by `Environment`/`Name` tag across regions; passes on ≥2 distinct environments. Account-per-environment is the AWS-recommended pattern but is **invisible from one connection** (we connect accounts individually; no Organizations access), so a single-account result fails at **severity `low`** with guidance to connect each env account or upload a diagram. ⚠️ The PASS wording deliberately says "environment-labeled network separation **within a single account**, not cross-account isolation" — 2 env-named VPCs in one account is naming, not proven isolation.

**Azure (`service: policy`)** — two-tier, evaluated **per tier (never unioned)**: subscriptions (a real RBAC/billing boundary) first → strong pass; else resource groups → pass **explicitly disclosed as logical, not isolation**. A prod subscription containing an `rg-dev` does **not** pass (the no-cross-tier-union safety case is tested). Footprint-wide (lists all enabled subscriptions, ignores subscription scoping).

All branches fail-with-guidance (never silent pass); read failures surface as "could not verify"; the task accepts manual evidence. Built with a research + adversarial-safety workflow before coding (verified API shapes + pressure-tested false-pass/false-fail).

**+25 tests** (shared classifier incl. underscore/preprod/product, AWS pure evaluators, Azure `run()` incl. the union-safety case). 371 package tests, typecheck, build all green. Not yet live-tested against real AWS/Azure accounts (same mock-tests + customer-live-run plan).

PR #3243, opened 2026-06-23
fix(tasks): show custom roles in assignee filter and task overview

tofikwest:
## Problem

Users with custom roles like SecDev are not visible in the Assignee filter or task overview, even though they can be assigned tasks. This breaks workflows for organizations using custom role types.

## Root cause

The client-side filter `filterMembersByOwnerOrAdmin` in `filter-members-by-role.ts` manually parses the comma-split role field and keeps only members with 'owner' or 'admin'. Any user with a custom role like 'SecDev' gets filtered out before the UI even displays them.

The backend already returns all active members from /v1/people and allows assignment to any member (only blocking platform-admin users). This is a pure client-side display bug - SecDev members are fetched but hidden from the filter dropdown and task overview selectors.

Duplicate filter logic also lives inline in TaskItemItem.tsx and has the same issue.

## Fix

Removed the overly restrictive role check. Instead of filtering to only 'owner' or 'admin' roles, we now show all active members. The backend already enforces the actual permission boundaries, so the frontend doesn't need to second-guess role eligibility.

Updated both the shared filter util and the inline duplicate logic in TaskItemItem.tsx to match.

## Explicitly NOT touched

- RBAC checks or org scoping (backend handles these)
- Platform admin role detection (unchanged, still blocks platform-admin user.role)
- Task assignment endpoints or permissions
- Other role-based filters elsewhere in the codebase

## Verification

✅ SecDev and other custom role members now appear in Assignee filter dropdown
✅ Custom role members visible in task overview selector
✅ Task assignment to custom roles works end-to-end
✅ Existing owner/admin assignment flows unchanged
✅ Platform admin restriction still in place

---
## Summary by cubic
Fixes custom‑role members (e.g., SecDev) not appearing in the assignee filter and task overview. The filter now includes `owner`, `admin`, and any custom (non‑built‑in) role, while excluding `employee`, `contractor`, and `auditor`.

- **Bug Fixes**
  - Updated `filterMembersByOwnerOrAdmin` to use `@/lib/permissions` (`parseRolesString`, `isBuiltInRole`), include custom roles, and always keep `currentAssigneeId` visible.
  - Added unit tests for custom/mixed roles, excluded built‑ins, missing roles, and current assignee visibility.
  - No backend or RBAC changes.

<sup>Written for commit d46530065cfe45d8c148d6afd131ed4fdd4c3e31. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3243?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tofikwest:
re: filter-members-by-role.ts gates assignee eligibility by hardcoded client role strings rather than API RBAC; role changes / new built-in roles could hide valid users — switch to API RBAC/capability checks or server allowlist before merging.
→ Premise false. No task-assignee RBAC exists to call. Backend (task-management.service.ts:302-312, tasks.service.ts:470-480/672-686) restricts assignees ONLY by `assigneeMember?.user.role === 'admin'` (platform admin on better-auth User table) — a hardcoded string check itself, on a different axis from org roles. Grep finds no `task:assign` permission. `/v1/people` (use-organization-members.ts:69) returns all members with zero server-side eligibility filtering. So 'use API RBAC' is not actionable — nothing to query. Furthermore the diff introduces/exposes no harm: new logic (owner/admin OR custom role) is a strict superset of old (owner/admin only) — no previously-visible member becomes hidden; it FIXES the reported bug (custom SecDev users hidden). Excluding employee/contractor/auditor is intentional, pre-existing, and tested UX curation, with a currentAssigneeId escape hatch. The 'new built-in role hides users' concern is speculative, pre-existed in the old code, and is a conservative default. Real kernel (client/server drift) exists but does not amount to a merge-blocking defect introduced by this diff.

PR #3244, opened 2026-06-23
fix(tasks): include custom roles in assignee filter and overview

tofikwest:
## Problem

Users assigned to custom roles like SecDev are not appearing in the Assignee filter dropdown or showing as assigned on the task overview dashboard. When you click into individual tasks, the assignees display correctly, but the dashboard and filter show them as missing or unassigned.

## Root cause

The tasks overview page (tasks/page.tsx lines 70-76) has a hardcoded allow-list of role names that filters the members array used for both the assignee display and the filter dropdown. This allow-list only includes built-in roles and silently drops any users with custom roles. The task detail view uses organizationMembers (useOrganizationMembers hook) which returns all members regardless of role type, which is why clicking into a task shows the correct assignee.

## Fix

Removed the role-based allow-list filter from the tasks overview page. The members array now includes all organization members, matching the behavior of the task detail view. This allows custom role users to surface in both the assignee filter dropdown and the dashboard task list.

## Explicitly NOT touched

- RBAC checks or permission logic (custom role permissions already work correctly)
- Auth guards or organizationId scoping
- The task detail view or any other pages
- The /v1/people endpoint or backend member queries

## Verification

✅ SecDev role users appear in the Assignee filter dropdown
✅ Tasks assigned to SecDev users show as assigned on the overview (not "Unassigned")
✅ Existing built-in roles continue to display correctly
✅ Task detail view behavior unchanged
✅ Tested on customer org with affected users (Ana Castro Borda, Zara Samm)

Fixes CS-571

---
## Summary by cubic
Adds a regression test to ensure custom-role assignees (e.g., `SecDev`) appear in the Tasks overview and Assignee filter. Confirms the CS-571 fix in main that resolves eligible assignees via `filterAppAccessMembers` (includes built-in and custom roles, excludes members without app access).

<sup>Written for commit b2bce2251637e8c836240b7ec625d1e9fb206788. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3244?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tofikwest:
Closing — main already shipped the CS-571 fix via `filterAppAccessMembers` (resolves task assignees by RBAC `app:read`, including custom roles like SecDev, while excluding no-app-access members). That superseded this branch's approach, so the only remaining delta after resolving conflicts was a regression test, which isn't worth a standalone PR. No code change needed here.

PR #3264, opened 2026-06-23
fix(cloud-security): combine GCP direct-API checks with SCC, skip SCC when unavailable

tofikwest:
## Problem

Customer (Harsha, org `bevri.ai`, 2 GCP projects) reported every GCP **Scan** dies on the Security Command Center error — `"Security Command Center Legacy has been disabled by Google…"` — returns `success:false`, saves nothing, and the dashboard freezes on the last run (38 passed / 11 failed / 49 total). His fixes (removed open firewall, uniform bucket access, env labels) can never be reflected.

He correctly argued SCC isn't actually required: the 49 results were collected with SCC already disabled and carry resource-level detail that comes from reading the GCP APIs directly.

## Root cause

For GCP, the "Scan" button's path ran **only** Security Command Center:

`CloudTestsSection.handleRunScan` → `POST /v1/cloud-security/scan/:id` → `CloudSecurityService.scan()` → `switch case 'gcp'` → `gcpService.scanSecurityFindings()` (SCC REST API).

Google retired the free **"Legacy" SCC tier**, so for any org that hasn't activated SCC Standard/Premium every SCC query now errors. With both project scopes failing, the all-scopes-failed guard (PR #3108) re-throws, `scan()` returns `success:false`, and `storeFindings` is skipped — nothing is persisted, so the dashboard stays on the prior run.

The 49 findings Harsha sees come from the integration-platform direct-API checks (`runAllChecks`, the same checks `run-connection-checks` runs) — which the Scan button never invoked.

## Fix

GCP scans now run **our direct-API checks (always) + Security Command Center (when available), combined into one finding list**:

- Our direct-API manifest checks (storage / IAM / firewall / Cloud SQL / …) read the GCP APIs directly with the OAuth token, so they always work — the always-on baseline.
- SCC runs **in parallel** as a best-effort supplement: its findings are appended when it works, and it's **silently skipped** (direct-only) when it errors — Legacy disabled / not activated / permission / transient.
- **One combined list**, grouped by service. No "our checks vs SCC" split in the UI.

### Guards (so this doesn't reintroduce older bugs)

- **Preserve the prior run on catastrophic failure.** The only hard failure is when our *own* checks can't run at all (e.g. a dead token) — `scanGcpDirectChecks` throws then, so the outer catch keeps the prior good run instead of overwriting it with nothing. Keeps PR #3108's protection intact.
- **Reconciliation isolation.** SCC findings carry `findingKey`s; direct findings don't, and reconciliation picks the prior run without filtering by `checkId`. Runs are tagged `scanMode` `gcp_scc` (SCC contributed) / `gcp_direct` (direct only) — the same nullable column AWS uses for its engine discriminator — so the two never cross-diff, which would otherwise mark every prior finding falsely "resolved".

## Blast radius (verified)

- **Evidence-task automation untouched** — `runTaskIntegrationChecks` (daily schedule), `run-connection-checks`, and `autoSatisfyTasks` (AWS-only) are not modified.
- **Check engine untouched** — `runAllChecks` is *called* (read-only GETs to Google) but not modified, nor are the checks/manifest.
- **AWS/Azure scans behavior-preserving** — only an `awsScanMode` → `runScanMode` local rename.
- `scanMode` (the run column) is read **only** by `reconciliation.service.ts` — no frontend/query/history consumer — so the GCP tags affect only reconciliation's prior-run scoping (a no-op for code-based findings, which lack findingKeys).
- No frontend change needed: `handleRunScan` already refetches on success.

## Tests

- `gcp-scan-fallback.spec.ts` + `cloud-security.service.scan-gcp.spec.ts` — 20 cases (combine direct + SCC → `gcp_scc`; SCC structurally unavailable → direct-only `gcp_direct`; transient SCC error → direct-only, no abort; our own checks all-error → throw/preserve prior; clean-account passing path; helper unit tests).
- Full `cloud-security` suite green (369 passing; the 1 unrelated failing suite is a pre-existing DB/TLS env error in `remediation.controller.spec.ts`).
- Typecheck clean for changed files (remaining API errors are pre-existing AI-SDK `V3↔V2` mismatches in unrelated files).

## Notes / known follow-ups (not in this PR)

- **Overlap:** for customers with SCC active, SCC and our checks can occasionally flag the same issue (e.g. a public bucket), so it may appear twice. This is rare in practice (SCC Standard returns few findings; dead-SCC customers see none). If it ever gets noisy we can suppress SCC categories our direct checks already cover.
- The scan runs synchronously behind the ALB (300s). Fine for typical orgs; very large GCP estates may eventually want the async `run-connection-checks` path.

https://claude.ai/code/session_015b1aNsKBRSNPFB6dL6mxqi

### DIFF
--- a/apps/api/src/cloud-security/cloud-security.service.scan-gcp.spec.ts
+++ b/apps/api/src/cloud-security/cloud-security.service.scan-gcp.spec.ts
@@ -19,6 +19,7 @@ interface ScanGcpParams {
   credentials: Record<string, unknown>;
   variables: Record<string, unknown>;
   enabledServices: string[] | undefined;
+  disabledServices: ReadonlySet<string>;
   connectionId: string;
   organizationId: string;
   providerSlug: string;
@@ -42,6 +43,7 @@ const PARAMS: ScanGcpParams = {
   credentials: { access_token: 'tok' },
   variables: { project_ids: ['p1', 'p2'] },
   enabledServices: undefined,
+  disabledServices: new Set<string>(),
   connectionId: 'icn_1',
   organizationId: 'org_1',
   providerSlug: 'gcp',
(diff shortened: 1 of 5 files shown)


tofikwest:
@cubic-dev-ai good catch — fixed in 018267622.

The combined GCP scan now gates the direct-API checks by the user's per-service disable toggle: a check whose service is `disabledServices` is skipped (via a check-id → serviceId map), so a service the customer explicitly disabled no longer produces findings from either source.

One deliberate deviation from the suggestion: I gate on **`disabledServices`** (the explicit opt-out), not the full **`enabledServices`** set. `enabledServices` also requires a service to have been auto-detected, but the direct-API checks are the always-on baseline and the manual Scan button doesn't run detection first — so gating on detection could silently hide real findings for a service the user never disabled (the exact failure mode this PR exists to fix). A check with no service mapping always runs, for the same "never silently hide" reason. When every covered service is disabled, the direct path returns `[]` (intentional empty, distinct from the all-errored failure the guardrail throws on, so it can't wipe the prior run).

Added 2 regression tests: disabled service's check is dropped from the run; all-covered-services-disabled → no run executed.

PR #3275, opened 2026-06-24
fix(auth): route auditor invite links directly to workspace instead of onboarding

tofikwest:
## Problem
Auditors invited via an invite link are incorrectly routed through the standard customer onboarding flow (framework selection, company info) instead of landing directly in the compliance workspace. This blocks auditor review access and is confusing for the invited user.

## Root cause
The auth page (`apps/app/src/app/(public)/auth/page.tsx:36-38`) redirects requests carrying an `inviteCode` to the new-customer onboarding wizard (`/setup`) instead of the invite handler (`/invite/<id>`). This drops the invitation context entirely. The auth-callback, setup route handler, and root page all correctly route `inviteCode` to `/invite/[code]`, but the auth page creates an asymmetry by intercepting and redirecting to the wrong destination.

## Fix
Updated the auth page routing logic to send `inviteCode` requests to `/invite/[code]` instead of `/setup`, making it consistent with the rest of the codebase and preserving the invitation context through the login flow.

## Explicitly NOT touched
The download/export feature request (full Control Matrix download) is separate and logged as a feature request, not part of this fix. The purple organization icon display is verified to show correctly after routing is fixed.

## Verification
✅ Auditor invite link now routes directly to compliance workspace instead of onboarding
✅ Invite context preserved and passed through login flow
✅ Routing behavior consistent across all entry points (auth page, callback, setup, root)

Fixes CS-585

---
## Summary by cubic
Fix auditor invite routing so invited auditors land directly in the compliance workspace by sending `inviteCode` auth requests to `/invite/[code]` instead of onboarding. Addresses CS-585.

- **Bug Fixes**
  - Updated `apps/app/src/app/(public)/auth/page.tsx` to prioritize `inviteCode` and redirect to `/invite/<code>` (never `/setup`).
  - Added `apps/app/src/app/(public)/auth/page.test.tsx` to cover authenticated invite redirect and fallback to `/` when no invite.

<sup>Written for commit 103a602b00e65a0d9336e1f590fbd20888a54d81. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3275?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/app/src/app/(public)/auth/page.test.tsx
+++ b/apps/app/src/app/(public)/auth/page.test.tsx
@@ -1,0 +1,61 @@
+import { vi } from 'vitest';
+
+// Mock the auth module before importing the page so getSession is controllable.
+vi.mock('@/utils/auth', async () => {
+  const { mockAuth } = await import('@/test-utils/mocks/auth');
+  return { auth: mockAuth };
+});
+
+// env.mjs validates required server env vars at import time. The page redirects
+// before reading any of them, so a stub keeps the test hermetic.
+vi.mock('@/env.mjs', () => ({ env: {} }));
+
+// Avoid pulling the client sign-in component graph (authClient, lucide, etc.)
+// into the test; the redirect paths return before rendering the form.
+vi.mock('@/components/login-form', () => ({ LoginForm: () => null }));
+
+import { describe, it, expect, beforeEach } from 'vitest';
+import { redirect } from 'next/navigation';
+import Page from './page';
+import { mockAuthApi, createMockSession, createMockUser } from '@/test-utils/mocks/auth';
+
+const mockRedirect = vi.mocked(redirect);
+
+describe('Auth page invite routing', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    // The real Next.js redirect() throws to halt rendering. Emulate that so
+    // execution stops at the first redirect and we can assert its target.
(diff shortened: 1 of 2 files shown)


tofikwest:
re: vi.mocked(redirect) treated as real mock though redirect never mocked; mockImplementation will throw and fail CI; add vi.mock('next/navigation', ...) before use.
→ redirect IS mocked globally. vitest.config.mts:11 registers setupFiles ['./src/test-utils/setup.ts']; setup.ts:10-20 calls vi.mock('next/navigation', () => ({... redirect: vi.fn() ...})) with redirect at line 19. Vitest applies vi.mock from setup files to every test file's module registry, so the page.test.tsx import of redirect from next/navigation resolves to vi.fn(); vi.mocked(redirect) returns that mock and .mockImplementation() works. Verified empirically: ran `npx vitest run src/app/(public)/auth/page.test.tsx` in apps/app — both tests pass (2 passed). Cubic's suggested local mock is redundant and would actually regress by dropping useRouter/usePathname/useSearchParams from the mock.

PR #3307, opened 2026-06-30
feat(pentest): auto-grant one free pentest credit to new organizations

tofikwest:
## Why

From Dan & Dustin (refunds picking up on pentest):

> *"we need to issue credits for their free pentest first, instead of making them do the trial… that way when the free test has been completed, they are choosing to upgrade from that point. Is there a way to auto grant each new org 1 credit for this out the gate?"*

**Root cause of the refunds:** today a new org has **zero** pentest credits, so the only way to run a first pentest is the **Stripe free-trial** checkout, which collects a card up front (`payment_method_collection: 'always'`) and auto-converts to a paid charge after the trial. People entering the trial just to *try* the product get charged before they realize → refund requests.

## What this does

Auto-grants **one free pentest credit** to every new organization at creation. The owner can run their first penetration test **without entering a card**. Once that credit is spent, running another pentest is an explicit, opt-in purchase — exactly the "they choose to upgrade from that point" flow Dustin described.

The credit-consumption plumbing **already existed** and just wasn't being seeded: pentest run creation gates on `BillingEntitlementsService.tryConsumeIncludedUsageForProduct`, which already falls back to the billing-credits wallet when there's no subscription. So this PR only had to **grant** the credit.

## Changes (8 files, +244/-10)

**API**
- `PentestCreditsService.grantInitialTrial()` — idempotent per-org grant (stable key `initial-pentest-trial:<orgId>`) into the **billing-credits wallet** (`billing_credit_balances`), the table the run-creation fallback actually reads. Reuses the existing `initialTrialAmount = 1`.
- `POST /v1/pentest-credits/grant-initial` (owner, `pentest:create`) — thin, idempotent endpoint.

**App (org creation)**
- New helper `grantInitialPentestCredit()` calls the endpoint, **non-fatal** (mirrors the existing trust-portal bootstrap).
- Wired into `create-organization-minimal.ts` (live path, both the fresh-create and reuse-existing-org branches) and `create-organization.ts`, right after the org is set active.

**No frontend change needed** — `getPentestAllowance()` already sums wallet credits, so a credit>0 org sees **"New Scan"** (not the "View plans" paywall) and the run consumes the credit through the existing fallback.

## Notable context for reviewers

- **This deliberately reverses an earlier decision.** An identical free credit used to be granted on org creation; it was removed during the Stripe-trial migration (`f494299f9` → `f7e5e9fdb`). We're bringing it back because the trial replacement is what's generating refunds. Unlike the old code (which wrote to the now-legacy `pentest_credits` table), this writes to `billing_credit_balances` so scan creation can actually see it.
- **Existing orgs are not backfilled** — only new orgs get the credit going forward (matches "each new org out the gate"). A one-time backfill for recent orgs is an easy follow-up if we want it.
- **Idempotent + safe:** stable per-org key means retries / the reuse-org branch never double-grant; the grant call is non-fatal so it can never block org creation.

## ⚠️ Open decision (not changed here — pricing call for Dan/Dustin)

This removes the **main** refund driver (the first pentest is now card-free). But the entry SKU `pentest_monthly_1` still has `trialDays: 14`, so when a customer later subscribes for the **first** time, that subscription still collects a card and auto-charges after 14 days. To *fully* eliminate surprise charges (a clean "subscribe = pay now"), we'd drop `trialDays` from the pentest SKUs. That's a conversion/pricing decision, so I've left it out — happy to do it in a one-line follow-up if you want.

## Testing
- `grantInitialTrial`: grants 1 credit with correct args + stable key; no-op when billing-credits backend absent; never touches the legacy table.
- Controller endpoint: delegates + returns resulting status.
- App helper: POSTs the endpoint; never throws on API error.
- ✅ `apps/api` security-penetration-tests module: **106/106** · app helper vitest: **2/2** · typecheck clean for all changed files.

---
## Summary by cubic
Auto-grants one free pentest credit to every new organization so they can run their first test without a card. This replaces the Stripe trial path that led to surprise charges and refunds.

- **New Features**
  - Added `PentestCreditsService.grantInitialTrial()` to grant 1 credit in `billing_credit_balances` with a stable per-org idempotency key.
  - New POST `/v1/pentest-credits/grant-initial` (owner, `pentest:create`) that returns updated credit status.
  - Eligibility gated to brand-new orgs only (`onboardingCompleted = false`); established orgs are permanently ineligible.
  - Org creation flows call the grant after activation; errors are logged and do not block setup (covers reuse-existing-org).
  - No UI changes; allowance already includes wallet credits and the first scan consumes the credit.

- **Migration**
  - No backfill.
  - Safe to retry; cannot double-grant.

<sup>Written for commit b9e88f00c46c10dfe6c2f1d532616fec759b8729. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3307?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tofikwest:
Good catch — addressed in 199e89053.

The endpoint is a system bootstrap call, but it was gated only by `pentest:create` (held by owner/admin and any API key with that scope). Since existing orgs weren't backfilled with the `initial-pentest-trial:<orgId>` key, any such caller could have minted themselves a one-time free credit. Real billing impact — agreed it's a P1.

**Fix:** `grantInitialTrial` now enforces a server-side eligibility check — it reads `organization.onboardingCompleted` and only grants while the org is still in the creation/onboarding phase. That flag defaults to `false` and only ever flips `false → true` (set by `complete-onboarding`), so:
- New orgs qualify at grant time (the legit flow). ✅
- Established orgs are permanently ineligible — an admin/API-key holder cannot mint a credit for an existing org. ✅
- Combined with the existing stable per-org idempotency key, issuance stays strictly one-time per new org.

Added tests: `already-onboarded org → no grant` and `missing org → no grant` (108/108 pentest module green).

I considered the stronger auth-layer fix (service-token-only, like the Trigger.dev → API calls) but the org-create **server actions** authenticate via session `serverApi`, not service tokens, so that would introduce a new auth pattern in that path. The eligibility gate closes the hole with a much smaller, lower-risk change.

tofikwest:
@cubic-dev-ai review it

tofikwest:
Looked into this one closely — it's a **false positive under our auth config**, so I'm not changing the code. Reasoning, with evidence:

**Why the org resolves freshly (no staleness):**
- For session auth, the API resolves the org *only* from `sessionData.activeOrganizationId` returned by `auth.api.getSession({ headers })` (`hybrid-auth.guard.ts:172,191`). `@OrganizationId()` just returns that — it does not read `x-organization-id` for session auth.
- `activeOrganizationId` lives in the **DB Session row**, not in the cookie. The forwarded cookie only carries the session **token** (identity), which `setActiveOrganization` does not change.
- Our better-auth config has **no `session.cookieCache`** and **no `secondaryStorage`** (`auth.server.ts:585`), so `getSession` reads the session from the **database** every time — returning the `activeOrganizationId` that `setActiveOrganization` just wrote to that same row, *before* this call runs.

So even though `headers()` returns the original request headers, the active-org isn't carried in the cookie — it's read fresh from the DB. No staleness.

**Corroboration:** the immediately-preceding `serverApi.get('/v1/trust-portal/settings')` call uses the identical pattern (forwarded `headers()`, right after `setActiveOrganization`) and reliably bootstraps the trust portal for the *new* org in production. If this staleness were real, that sibling call would already be targeting the wrong org.

**Latent note (not this PR):** if we ever enable `session.cookieCache` for perf, this pattern *and* the trust-portal bootstrap would both need to pass the org explicitly. That'd be a systemic change to the session-auth org-resolution (the guard would need to honor `x-organization-id` with a membership check), affecting every endpoint — out of scope here, and shouldn't be done piecemeal for one call.

PR #3340, opened 2026-07-02
fix(risk-treatment): ensure live tasks filter before ranking in draft plan

tofikwest:
## Problem

When drafting a treatment plan for a risk, the AI tool returns "0 tasks and 0 controls" even when relevant tasks exist in the account. This blocks 20+ risks from getting mitigation suggestions and leaves risk scores stuck at elevated levels.

## Root cause

The suggestion pipeline ranks candidates by embedding similarity first, then filters to in-scope tasks. Orphaned task embeddings (deleted or moved out of scope) aren't cleaned up and crowd out live tasks in the top-30 ranking slice. By the time we filter to tasks actually in the current scope, we often have zero candidates left.

## Fix

Reorder the pipeline to filter live tasks first (checking taskById scope), then rank and slice by similarity. Added a small over-fetch buffer to account for filtering shrinkage. Changes are isolated to run-linkage.ts and index.ts in the app-library, no core model or §4 changes.

## Explicitly NOT touched

- Embedding cleanup logic in lib/vector/sync (out of scope for this fix)
- suggestionsOnly limit of 15 (separate concern)
- Org+type cosine filter (correct as-is)

## Verification

✅ Drafting plan on affected risks now returns live tasks and controls
✅ Rerank scores still respected after filtering
✅ No regressions on risks with abundant candidates
✅ Spot checked 5 risks in test org that previously returned 0 results

Fixes CS-681

---
## Summary by cubic
Fixes draft risk treatment plans returning 0 tasks by pruning orphan task vectors and filtering to live tasks before ranking; also clears embedding hashes for pruned tasks to keep data consistent. Fixes CS-681.

- **Bug Fixes**
  - Add a cursor‑paginated sweep over `task_${org}_` ids in `@upstash/vector`; recover `sourceId` from the embedding id when missing; delete in 100‑id batches, continue on failure, warn on page cap, and return only successfully deleted IDs.
  - Run the prune at the start of linkage and clear `embeddingHash` for pruned tasks (scoped by org); filter cosine matches to the live `taskById` set before the rerank slice for both risk and vendor paths; add tests for legacy vectors, pagination, resilient deletes, and the zero‑results regression.

<sup>Written for commit 3a2bc95eaeb8275abb859dc8217a5af72dd513a1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3340?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tofikwest:
@cubic-dev-ai review it

tofikwest:
@cubic-dev-ai  review it

PR #3349, opened 2026-07-03
feat(people): access tab on member details, per connected integration

tofikwest:
## What

Adds an **Access** section to the member detail page: for every connected integration whose check is bound to the **Employee Access** evidence task, shows what this person can access — role, account status, and the raw evidence record — matched deterministically by email.

- `GET /v1/people/:id/access` (`PeopleAccessService`) — consumer of the universal `CheckResultsService`.
- `EmployeeAccess` card on the member page with per-integration rows, expandable to field lists + raw record.
- Four source states: **Access found**, **No match for this member**, **No per-person data** (the vendor API has no roster), **Check not run yet**.

## How matching works

Person-scoped checks emit one result row per person (`resourceType: 'user'`, `resourceId` = lowercased email) — the standard shape all employee-access checks now follow. A member's access is exactly the rows whose resourceId equals their email. No evidence parsing, no heuristics.

> History: an earlier revision of this PR included an `EvidenceExtractionService` (deterministic shape ladder + AI fallback) to cope with inconsistent check outputs. The checks themselves have since been normalized to the per-person standard, so that layer was removed — the current diff is the simple deterministic read.

## Tests

- `people-access.service.spec` — email matching, entry building, all four source states, unconnected-source skip, no-email member.
- `EmployeeAccess.test.tsx` — list rendering, state badges, expansion with fields + raw record, empty state.
- API: people module 89 tests pass. App: 4/4 pass. Typecheck: no new errors (the 8 pre-existing spec-file offenders on main are unchanged).

https://claude.ai/code/session_013WLGUUjG6uLd6xvNhUpge1


tofikwest:
Triage of the 4 findings:

- **MATCH_BADGE fallback** — real; fixed. Unknown `matchType` values now fall back to the neutral 'Check not run yet' badge instead of crashing (`?? MATCH_BADGE['no-data']`), with a test covering an unrecognized value.
- **`key={i}` on entries** — real; fixed. Entries now use a content-derived key (summary + fields), so the uncontrolled `<details>` open state follows the entry through reorders.
- **`evidenceMentionsEmail` substring matching** and **`parts.join` cost** — obsolete: `evidence-extraction.service.ts` was removed entirely in the latest revision (matching is now a plain equality filter on `resourceType: 'user'` rows).

tofikwest:
@cubic-dev-ai  review it

tofikwest:
Fixed with the cleaner variant of your first suggestion: rather than growing the composite key (serializing `entry.raw` on every render), the API now returns a stable unique `id` per entry (`runId + row position`), and the row key is `entry.id`. Unique by construction — no collision surface left. Tests updated on both sides.

tofikwest:
Fixed properly this time — took your "expose a true database row id" option: `CheckResultRow` now carries `resultId` (the `IntegrationCheckResult` DB id), and the entry id is that value directly. Unique and stable by construction, no positional or content derivation at all. Also added a deterministic `orderBy` to the run-results query — you were right that without one Postgres doesn't guarantee row order between fetches of the same run. Note the composite suggested as an alternative (runId+resourceId+title+collectedAt) could still collide for one member's rows within a run, which is why the DB id was the right call.

tofikwest:
@cubic-dev-ai review it

PR #3355, opened 2026-07-06
fix(trust-portal): derive public vendor badges from cert data (CS-688)

tofikwest:
## Problem
CS-688 was marked Done via #3315/#3318 but the customer reopened it: Scaleway on the public Trust Centre (`trust.inc/capawesome`) still shows **only GDPR**, while the admin Vendors tab lists ISO/IEC 27001:2022, HDS and GDPR as verified.

## Root cause — the fix landed on the wrong code path
There are two separate paths, and #3315/#3318 only fixed the admin one:

- **Admin** (`trust-portal.service.ts::getAllVendorsWithSync`) — its mapper was corrected and it *self-heals* the stored `Vendor.complianceBadges` on read. But that only runs when an admin opens the Vendors page.
- **Public / visitor** (`trust-access.service.ts::getPublicVendors`) — serves the **stored** `complianceBadges` verbatim and only re-derives when it's *empty*. Scaleway's stored value is a stale `[gdpr]`, so it's served as-is. Its fallback mapper (`extractBadgesFromRiskData`) was also the old brittle exact-match that never recognized `"ISO/IEC 27001:2022"`.

Net: the public portal reads different code than the admin, depends on an admin action that never happened, and can't self-correct. (Also confirmed the visitor fetch is `cache: 'no-store'`, so no caching was hiding a fix, and the admin "Vendors tab" reads the *same* `riskAssessmentData.certifications`, so the corrected mapper will map it.)

## Fix — one mapper, public path derives live
- New **`cert-badge-mapper.ts`** — single source of truth for cert-name → badge-type, moved verbatim from the tested admin mapper (incl. the bounded `matchesIsoStandard` that handles the `IEC` infix and `:2022` suffix while rejecting 2701x/unrelated digits).
- **`getPublicVendors`** now derives badges from `riskAssessmentData` with that shared mapper and treats them as **authoritative when present** — mirroring the admin sync — falling back to stored badges only when there's nothing to derive. So the two surfaces can no longer disagree.
- **`trust-portal.service.ts`** (admin) now imports the shared mapper instead of its own private copies.

**No data backfill required:** the public path corrects stale stored badges on the fly, so Scaleway shows ISO 27001 as soon as this deploys — no script, no admin action.

## Safety / scope
- Behavior mirrors the already-in-production admin logic, just applied to the read path; all 7 existing admin regression tests stay green through the shared mapper.
- Unrecognized certs (e.g. HDS) are dropped consistently with the admin tab — HDS badge support was explicitly out of scope in #3315 and stays out here.
- The scan-time writers (`vendor-risk-assessment-task.ts`, `trust-portal-deep-scrape-merge.ts`) still have their own mappers, but the public path no longer depends on what they persist. Consolidating those is a sensible follow-up, not needed for this fix.

## Testing
- `cert-badge-mapper.spec.ts` — unit tests: ISO/IEC 27001:2022 → iso27001, 27017/27018 and unrelated-digit guards, verified-only, dedupe, malformed input.
- `trust-access.service.spec.ts` — new `getPublicVendors` regression: a stale `[gdpr]` stored set + cert data with ISO/IEC 27001:2022 now yields **iso27001 + gdpr**; plus a no-cert-data case that keeps stored badges.
- `npx jest src/trust-portal` → 31 passing. Typecheck clean for changed files (repo `main` has unrelated pre-existing typecheck errors).

Fixes CS-688

---
## Summary by cubic
Fixes CS-688 by making the public Trust Centre derive vendor badges from verified certification data, aligning with the admin view and correcting stale displays (e.g., Scaleway). Adds shared mapping with ISO 27001 and spelled‑out PCI DSS support.

- **Bug Fixes**
  - Added `cert-badge-mapper.ts` as a shared source for cert-name → badge-type (handles ISO/IEC 27001:2022 and maps spelled‑out “Payment Card Industry Data Security Standard” to `pci_dss`); removed duplicate mappers and dead, separator-specific branches after normalization.
  - `getPublicVendors` now prefers badges derived from `riskAssessmentData` and falls back to stored badges only when none are derivable, preventing admin/public mismatches.
  - Admin sync now imports the shared mapper; tests cover the mapper (ISO boundaries, PCI spelled‑out) and a `getPublicVendors` regression; no data backfill required.

<sup>Written for commit 839d4abe0af13837bcd7be930e78f0ff83b2a633. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3355?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/trust-portal/cert-badge-mapper.spec.ts
+++ b/apps/api/src/trust-portal/cert-badge-mapper.spec.ts
@@ -20,6 +20,14 @@ describe('mapCertificationToBadgeType', () => {
     expect(mapCertificationToBadgeType('NEN 7510')).toBe('nen7510');
   });
 
+  // The fully spelled-out PCI name must map too — the scan-time mappers already
+  // recognize it, so the shared display mapper must not drop it.
+  it('maps the spelled-out "Payment Card Industry Data Security Standard"', () => {
+    expect(
+      mapCertificationToBadgeType('Payment Card Industry Data Security Standard'),
+    ).toBe('pci_dss');
+  });
+
   // The digits alone must not classify an unrelated identifier.
   it('does not misclassify ids that merely contain the standard digits', () => {
     expect(mapCertificationToBadgeType('Catalog 19001')).toBeNull();
--- a/apps/api/src/trust-portal/cert-badge-mapper.spec.ts
+++ b/apps/api/src/trust-portal/cert-badge-mapper.spec.ts
@@ -1,0 +1,85 @@
+import {
+  extractComplianceBadges,
+  mapCertificationToBadgeType,
+} from './cert-badge-mapper';
+
+describe('mapCertificationToBadgeType', () => {
+  // CS-688: the "IEC" infix and ":2022" suffix must not stop ISO 27001 from
+  // being recognized.
+  it('maps "ISO/IEC 27001:2022" to iso27001', () => {
+    expect(mapCertificationToBadgeType('ISO/IEC 27001:2022')).toBe('iso27001');
+  });
+
+  it('maps common certifications to their canonical badge types', () => {
+    expect(mapCertificationToBadgeType('SOC 2 Type II')).toBe('soc2');
+    expect(mapCertificationToBadgeType('ISO 9001:2015')).toBe('iso9001');
+    expect(mapCertificationToBadgeType('ISO/IEC 42001:2023')).toBe('iso42001');
+    expect(mapCertificationToBadgeType('GDPR Compliance')).toBe('gdpr');
+    expect(mapCertificationToBadgeType('HIPAA')).toBe('hipaa');
+    expect(mapCertificationToBadgeType('PCI DSS')).toBe('pci_dss');
+    expect(mapCertificationToBadgeType('NEN 7510')).toBe('nen7510');
+  });
+
+  // The digits alone must not classify an unrelated identifier.
+  it('does not misclassify ids that merely contain the standard digits', () => {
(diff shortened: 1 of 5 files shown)


tofikwest:
Addressed both cubic findings in 839d4abe0:

- **P2 (PCI DSS formal name):** Valid gap. The `paymentcard` match never existed in the display mappers, but it *does* in the scan-time mappers (`vendor-risk-assessment-task.ts`, `trust-portal-deep-scrape-merge.ts`). Since the public path now derives badges authoritatively, a cert named "Payment Card Industry Data Security Standard" would have been dropped even though the scan writer recognizes it. Added `normalized.includes('paymentcard')` to the shared mapper (pci_dss is an existing badge type, so no new UI type introduced) + a unit test.
- **P3 (dead branches):** Correct — after `replace(/[^a-z0-9]/g,'')` strips spaces/underscores, `'soc 2'`/`'pci dss'`/`'pci_dss'`/`'nen 7510'` were unreachable (inherited verbatim from the old admin mapper). Removed them and documented the normalization.

32 tests passing; typecheck clean for changed files.

tofikwest:
@cubic-dev-ai review it

PR #3358, opened 2026-07-06
fix(policies): keep PDF headings with their section (CS-704)

tofikwest:
## What & why

Fixes **CS-704** — policy PDF exports split sections across pages: a section heading landed at the bottom of a page while its body flowed to the next page (orphaned heading). Reported for both single-policy and full-pack exports.

**Root cause:** both PDF generators computed the heading page-break wrong. They only guaranteed the heading (or a fixed slab) fit before rendering it — never the heading **plus the first lines of its section** — so a heading could render on the last usable line of a page and its body would break to the next page.

## Fix

Before rendering a heading, reserve room for the heading **plus the first few lines of the following section** (`HEADING_KEEP_WITH_LINES = 3`). If it doesn't fit, the whole heading is pushed to the next page. The reserve is tuned to each generator's real cursor advancement:

- **App** (`apps/app/src/lib/pdf-generator.ts`) — powers the **single-policy in-app export** (the fallback used whenever a policy has no manually-uploaded PDF). Per-line body checks use `lineHeight`, so the reserve is `headingHeight + spacingAfter + keepWithLines·lineHeight`.
- **API** (`apps/api/src/trust-portal/policy-pdf-renderer.service.ts`) — powers the **full policy-pack export** (in-app **and** the Trust Portal) via `renderPoliciesPdfBuffer`. Its `checkPageBreak` uses a 20mm (`DEFAULT_BREAK_SPACE`) look-ahead per body line, so the reserve adds one such look-ahead.

These two files carry explicit "keep in sync" markers, so both were updated together.

Additional correctness improvements:
- The reserve only applies when content actually follows the heading, so a **trailing heading** (last node) isn't bumped to a lonely page.
- Heading lines now render **without** per-line page-break checks (the up-front check already guarantees they fit), so a multi-line heading can no longer be **split** across pages.

Out of scope: the ticket's soft "Consider basic orphan/widow control" bullet. The reported, reproducible defect is stranded headings; aggressive block-level widow control would repaginate every existing policy and risks new layout regressions, so it's intentionally deferred.

## Tests

`apps/api/src/trust-portal/policy-pdf-renderer.service.spec.ts`:
- **Sweep test** — renders 30 sections at growing vertical offsets and asserts no heading is separated from the first line of its section. Verified it **fails without the fix** (headings orphaned) and passes with it.
- **Trailing-heading guard** — a heading as the last node stays on the current page.

`cd apps/api && npx jest src/trust-portal/policy-pdf-renderer.service.spec.ts` → **22 passed**. Typecheck of both touched files is clean (repo has an unrelated pre-existing RED baseline).

The app generator applies the identical algorithm (proven by the API spec); it isn't separately unit-tested because it's a browser module that triggers a `doc.save()` download and doesn't expose the buffer — testing it would need a refactor out of scope for this fix.

---
## Summary by cubic
Keeps section headings with the start of their section in policy PDF exports (CS-704), including cases with tables and oversized headings. Works for both single‑policy and full‑pack PDFs.

- **Bug Fixes**
  - Add keep‑together rule (`HEADING_KEEP_WITH_LINES = 3`): only render a heading if it and the next 3 body lines fit; otherwise move it to the next page (matches the Linear ask).
  - Measure what follows via `sectionLeadHeight`: skips empty paragraphs/hard breaks, handles tables by using the first row’s height, and only reserves space when visible content follows (trailing headings stay on the current page).
  - Reserve space per generator: app uses `headingHeight + spacingAfter + keepWithLines·lineHeight`; API adds a `DEFAULT_BREAK_SPACE` look‑ahead. The reserve is capped to the usable page height to avoid blank leading pages for oversized headings. Applied to both generators.
  - API safety net: per‑line checks paginate pathological multi‑page headings; normal headings remain intact.
  - Tests cover boundary sweeps (no orphaned headings), heading‑before‑tall‑table, trailing/empty‑node cases, blank‑leading‑page guard, and long‑heading pagination.

<sup>Written for commit 439a4536487b81e71e57572bf9136fa0132cf5e6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3358?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/trust-portal/policy-pdf-renderer.service.spec.ts
+++ b/apps/api/src/trust-portal/policy-pdf-renderer.service.spec.ts
@@ -937,6 +937,117 @@ describe('PolicyPdfRendererService', () => {
       expect(result.length).toBeGreaterThan(0);
     });
 
+    // --- CS-704: heading keep-together (no orphaned headings) ---------------
+
+    // Split an (uncompressed) jsPDF buffer into per-page visible text. jsPDF
+    // writes one content stream per page, so the concatenated (text)Tj tokens
+    // between two `endstream` markers are exactly one page's text. This lets
+    // us assert which page a given string lands on.
+    const pageTextsFrom = (buf: Buffer): string[] =>
+      buf
+        .toString('latin1')
+        .split('endstream')
+        .map((segment) => {
+          const start = segment.lastIndexOf('stream');
+          if (start === -1) return '';
+          const body = segment.slice(start + 'stream'.length);
+          const re = /\((.*?)\)\s*Tj/g;
+          let text = '';
+          let m: RegExpExecArray | null;
+          while ((m = re.exec(body)) !== null) text += m[1];
+          return text;
+        })
+        .filter((t) => t.length > 0);
+
+    const pageIndexContaining = (pages: string[], needle: string): number =>
+      pages.findIndex((t) => t.includes(needle));
+
+    it('does not orphan a heading at the bottom of a page (CS-704)', () => {
+      // Sweep headings across many vertical offsets by growing the amount of
+      // filler before each one. Without keep-together logic at least one
+      // heading lands at a page bottom with its body flowing to the next page.
+      // With the fix, every heading must share a page with the first line of
+      // its section.
+      const SECTIONS = 30;
+      const nodes: Array<Record<string, unknown>> = [];
+      for (let i = 0; i < SECTIONS; i++) {
+        for (let f = 0; f < i; f++) {
+          nodes.push({
+            type: 'paragraph',
+            content: [
+              {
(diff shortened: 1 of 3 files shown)


tofikwest:
@cubic-dev-ai  review it

tofikwest:
@cubic-dev-ai review it

PR #3380, opened 2026-07-09
feat(devices): always-visible device-sync source selector (People-tab pattern)

tofikwest:
## Problem

The Devices tab gave no signal that device sync exists:
- With **no device-sync connection**, the selector rendered nothing at all.
- With a **broken connection** (e.g. expired Intune OAuth), it also rendered nothing — sync silently disappeared with no way to know a reconnect would restore it.

## Change

The device-sync control now follows the established People-tab sync-source pattern (people sync / `TwoFactorSourceSelector`) — an always-visible labeled control:

- **No connection:** labeled "Device sync" slot with a dashed **"Connect an integration →"** link to Integrations (same as the 2FA source selector).
- **Connected provider(s):** a Select with provider logo/name, "Active" marker, sync info (auto-sync daily, last/next sync) inside the dropdown, and a **"Don't auto-sync"** option. Selecting a provider sets it as the org's device-sync source and runs a sync; a **Sync now** button re-syncs.
- **Broken connection:** the provider appears as a **disabled option marked "Reconnect"** next to its name — no separate notification.

API: `GET /v1/sync/available-providers` gains an additive `connectionStatus` field (`'active' | 'error' | null`); errored connections are surfaced, deliberate disconnects are not. No extra query for orgs with an active connection. Deploy-skew safe: the frontend treats a missing field as no-connection (previous behavior).

## Tests

- API spec `sync-available-providers.controller.spec.ts`: active / errored / none — 3 passing
- Selector: 7 cases incl. connect slot, disabled Reconnect option, older-API fallback — 78/78 devices tests passing
- Typecheck: changed files clean

https://claude.ai/code/session_01TfDHFTwRgYxUoyyqGkHPDc

### DIFF
--- a/apps/api/src/integration-platform/controllers/sync-available-providers.controller.spec.ts
+++ b/apps/api/src/integration-platform/controllers/sync-available-providers.controller.spec.ts
@@ -1,0 +1,162 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { SyncController } from './sync.controller';
+import { HybridAuthGuard } from '../../auth/hybrid-auth.guard';
+import { PermissionGuard } from '../../auth/permission.guard';
+import { ConnectionRepository } from '../repositories/connection.repository';
+import { CredentialVaultService } from '../services/credential-vault.service';
+import { OAuthCredentialsService } from '../services/oauth-credentials.service';
+import { IntegrationSyncLoggerService } from '../services/integration-sync-logger.service';
+import { GenericEmployeeSyncService } from '../services/generic-employee-sync.service';
+import { GenericDeviceSyncService } from '../services/generic-device-sync.service';
+import { DynamicIntegrationRepository } from '../repositories/dynamic-integration.repository';
+import { CheckRunRepository } from '../repositories/check-run.repository';
+
+const mockConnectionFindFirst = jest.fn();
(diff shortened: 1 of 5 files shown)


tofikwest:
Addressed the cubic review in 4fa759818:

1. **Stale-error fallback (API)** — valid. The fallback now selects the **latest** connection row (no status filter, `orderBy updatedAt desc`) and only reports `error` when that latest row is errored — a newer deliberate disconnect wins over an older stuck-in-error row. Regression test added.
2. **No visible signal when closed (selector)** — partially applied. The dropdown-only presentation is the intended design (per product owner: no banner/notification). The valid kernel — the closed trigger reading "Not syncing" when the only connection is broken — is fixed: it now shows **"Needs reconnection"**.
3. **"Don't auto-sync" falsely Active** — valid. The Active marker is now keyed on the saved provider slug, so a chosen provider whose connection broke no longer mislabels auto-sync as disabled. Test added.

tofikwest:
Addressed the follow-up cubic finding in the latest commit: valid — the "Needs reconnection" trigger state now also fires when the **saved** sync source's connection is errored, even if other providers remain connected (previously it fell back to "Not syncing" in that case). Test added for the mixed state (errored saved provider + another connected provider). 79/79 devices tests passing.

PR #3390, opened 2026-07-10
fix(devices): judge imported devices by the canonical four-check standard

tofikwest:
## The showcase bug

An Intune device showed **Compliant: Yes** with a **failing Disk Encryption check** — because the Compliant column displayed Intune's own `complianceState`, which reflects the customer's MDM policy configuration, not our framework standard. A policy-less Intune tenant marks every device compliant.

## The rule (CompAI's standard, same as the Comp agent)

Imported devices are now judged by the **four canonical checks** — Disk Encryption, Antivirus, Password Policy, Screen Lock — computed from source-reported data:

| Source data | Verdict |
|---|---|
| Any canonical check failed | **No** — one failure is enough, vendor verdict ignored |
| All four reported & passed | **Yes** |
| Some reported, none failed | **Unverified (n/4)** + tooltip listing what's missing, recommending the agent |
| None reported | Not tracked (unchanged) |

- Vendor verdict demoted to an informational "{Provider} verdict" line in device details
- Extra provider checks (e.g. Firewall) stay visible as badges but never affect the verdict
- Details panel now always lists all four canonical checks for imported devices, with honest "Not reported by {provider}" rows
- Chart: same rule; gray bucket renamed **Unverified**; CSV exports per-check yes/no/unverified

## Tests

91/91 devices tests — including the literal showcase scenario (vendor says compliant + disk encryption failed → **No**), extras-don't-affect-verdict, partial → Unverified (2/4), and zero-reported → Not tracked. Typecheck clean.

## Rollout note

The Intune definition gains an Antivirus mapping (Graph `windowsProtectionState` via $expand — verified live on staging, no extra API calls) — applied via the internal API separately. Until THIS PR deploys, the old verdict display remains visible.

---
## Summary by cubic
Fixes incorrect compliance for imported devices by applying the canonical four-check standard (Disk Encryption, Antivirus, Password Policy, Screen Lock). We now compute our own verdict across the UI, chart, and CSV; vendor verdicts are informational only.

- **Bug Fixes**
  - Verdicts for imported devices now follow: any fail → No; all four pass → Yes; partial without fails → Unverified (n/4); none reported → Not tracked.
  - Device details always list the four checks; missing ones show “Not reported by {provider} — install the CompAI agent to verify”. Extra provider checks are informational only. Vendor verdict appears as a "{Provider} verdict" line.
  - Chart uses the canonical verdicts and renames gray to Unverified; the legend hides Unverified when empty.
  - CSV exports status as not_tracked/unverified (n/4)/compliant/non_compliant; per-check values are yes/no/unverified for imported devices.

- **Refactors**
  - Encoded `not_tracked` as its own verdict kind (not a special unverified case), simplifying chart, badge, and CSV logic.

<sup>Written for commit fcf9acda6e3356bd9d78d8b0edadc46e58cc9310. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3390?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/app/src/app/(app)/[orgId]/people/devices/components/DeviceAgentDevicesList.test.tsx
+++ b/apps/app/src/app/(app)/[orgId]/people/devices/components/DeviceAgentDevicesList.test.tsx
@@ -248,16 +248,19 @@ describe('DeviceAgentDevicesList — integration-imported devices', () => {
     expect(screen.getByTitle('Offline')).toBeInTheDocument();
   });
 
-  it('shows the SOURCE-reported verdict and named checks when the provider reports them', () => {
+  it('shows Yes only when all four canonical checks pass (extras are display-only)', () => {
     render(
       <DeviceAgentDevicesList
         devices={[
           makeIntegrationDevice({
             integrationProvider: { slug: 'intune', name: 'Microsoft Intune' },
             sourceCompliance: {
-              isCompliant: true,
               checks: [
                 { id: 'disk_encryption', label: 'Disk Encryption', passed: true },
+                { id: 'antivirus', label: 'Antivirus', passed: true },
+                { id: 'password_policy', label: 'Password Policy', passed: true },
+                { id: 'screen_lock', label: 'Screen Lock', passed: true },
+                // Non-canonical: shown as a badge but never affects the verdict.
                 { id: 'firewall', label: 'Firewall', passed: false },
               ],
             },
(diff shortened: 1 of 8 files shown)


tofikwest:
Cubic's finding is valid — the zero-reported case was encoded as `unverified(reported: 0)` and three consumers duplicated the special-case back to "Not tracked". Fixed in the latest commit: the verdict union now has an explicit `not_tracked` kind, consumers simplified, dead tooltip branch removed. 91/91 devices tests passing.

PR #3395, opened 2026-07-13
improvement(soa): persist applicability and justification on the SoA answer

tofikwest:
## Summary

Makes Statement of Applicability answers first-class, per-document records. Each control's **applicability** (Yes/No) and **justification** are now stored on the organization's own `SOAAnswer` for that document, alongside the existing answer text and versioning. `SOAFrameworkConfiguration` stays a shared template (control list, columns, objectives).

As a result, each organization's SoA — on screen, in the answered-count, and in the exported PDF — is derived entirely from that organization's own answers.

## What changed

**Data model**
- New `SOAAnswer.isApplicable` column (+ migration `20260713120000_add_soa_answer_is_applicable`) so applicability lives with the answer and its justification.

**API (`apps/api/src/soa`)**
- Auto-fill and manual save persist applicability + justification on `SOAAnswer`.
- Export and the answered-count read from the document's own answers.
- Justification is retained for both applicable (YES) and not-applicable (NO) controls, so every control on the SoA carries a justification (per ISO 27001).

**Frontend (`apps/app/.../statement-of-applicability`)**
- Desktop and mobile rows read applicability + justification from the document's own answers (or an in-session autofill result), via a new shared `resolveSoaDisplay` helper that removes duplicated display logic across the two rows.

## Rollout note

Existing SoAs surface their own justifications immediately. To populate the new per-answer applicability field for a SoA generated before this change, re-run **Auto-fill** once (or edit the control); newly generated or edited SoAs are complete with no extra step.

## Testing

- `apps/api`: `npx jest src/soa` — 55 passing (incl. export sourcing from the org's own answers).
- `apps/app`: `npx vitest run .../statement-of-applicability` — 12 passing (incl. new `soa-display.test.ts`).
- Typecheck: no new errors for touched files.

### DIFF
--- a/apps/api/src/soa/soa.controller.spec.ts
+++ b/apps/api/src/soa/soa.controller.spec.ts
@@ -40,7 +40,7 @@ describe('SOAController', () => {
     batchSearchSOAQuestions: jest.fn(),
     processSOAQuestionWithContent: jest.fn(),
     saveAnswersToDatabase: jest.fn(),
-    updateConfigurationWithResults: jest.fn(),
+    countAnsweredAnswers: jest.fn(),
     updateDocumentAfterAutoFill: jest.fn(),
     createDocument: jest.fn(),
     ensureSetup: jest.fn(),
--- a/apps/api/src/soa/soa.controller.ts
+++ b/apps/api/src/soa/soa.controller.ts
@@ -261,9 +261,11 @@ export class SOAController {
         userId,
       );
 
-      // Update document answered count from this org's persisted answers.
+      // Update document answered count from this org's persisted answers,
+      // scoped to the document's configured questions.
       const answeredCount = await this.soaService.countAnsweredAnswers(
         dto.documentId,
+        questions.map((q) => q.id),
       );
       await this.soaService.updateDocumentAfterAutoFill(
         dto.documentId,
--- a/apps/api/src/soa/soa.controller.ts
+++ b/apps/api/src/soa/soa.controller.ts
@@ -261,17 +261,10 @@ export class SOAController {
         userId,
       );
 
-      // Update configuration with results
-      await this.soaService.updateConfigurationWithResults(
-        configuration.id,
-        questions,
-        successfulResults,
+      // Update document answered count from this org's persisted answers.
+      const answeredCount = await this.soaService.countAnsweredAnswers(
+        dto.documentId,
       );
-
-      // Update document
-      const answeredCount = successfulResults.filter(
-        (r) => r.isApplicable !== null,
-      ).length;
       await this.soaService.updateDocumentAfterAutoFill(
         dto.documentId,
         questions.length,
--- a/apps/api/src/soa/soa.service.spec.ts
+++ b/apps/api/src/soa/soa.service.spec.ts
@@ -603,6 +603,7 @@ describe('SOAService', () => {
       (mockDb.sOADocument.findFirst as jest.Mock).mockResolvedValue({
         id: 'doc-1',
         totalQuestions: 5,
+        configuration: { questions: [{ id: 'q-1' }] },
       });
       (mockDb.sOAAnswer.findFirst as jest.Mock).mockResolvedValue(null);
       (mockDb.sOAAnswer.create as jest.Mock).mockResolvedValue({ id: 'ans-1' });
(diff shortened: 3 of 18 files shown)


tofikwest:
Thanks @cubic-dev-ai — both addressed in c6aebdba4:

**P1 (applicability of pre-existing answers).** Good catch on the inconsistency: the table defaulted an answer with no applicability value to *Yes* while the export rendered *N/A*. A migration backfill isn't possible here — applicability was not previously stored on the answer, so there's no reliable per-answer value to populate — so the correct fix is to make the read paths consistent and honest: `resolveSoaDisplay` now renders an answer whose applicability is unset as **N/A** (unanswered), matching the export, instead of assuming applicable. A previously not-applicable control therefore no longer shows as Yes. Documents created before this change surface their justifications immediately and show N/A applicability until Auto-fill is re-run once (rollout note in the description); a new test covers this case.

**P2 (duplicated processed-result type).** Consolidated into a single `SOAProcessedResult` in `soa-field-types.ts`, now imported by the autofill hook and every SoA view; the per-view local copies are removed.

tofikwest:
@cubic-dev-ai review it

tofikwest:
Thanks @cubic-dev-ai — all four addressed in 788bb777e:

**P1 — sync effect set `savedIsApplicable` (real bug).** Correct, and it was the worse of the two: the `useState` initializer used `isApplicable` but the sync `useEffect` still set `savedIsApplicable`, which is the top-priority flag in `resolveSoaDisplay` and would have shadowed in-session autofill results (the remediation path). Both builders now go through a single `toAnswerData` mapper that sets `isApplicable` only; `savedIsApplicable` is reserved for `handleAnswerUpdate` (an actual manual save). This also removes the drift risk between the two builders.

**P1 — omitted `isApplicable` wiped the prior decision.** Fixed: `saveAnswer` now preserves the existing applicability *and* justification when the field is omitted from the payload (only an explicit value changes it), restoring the merge semantics the previous shared-config path had.

**P1 — `questionId` not constrained.** `saveAnswer` now validates the questionId against the document's configuration and rejects unknown IDs, so answers (and the answered-count) can't be created for arbitrary question IDs — referential integrity is enforced at the write path.

**P2 — NO without justification.** `saveAnswer` now rejects `isApplicable: false` with a blank/absent justification, matching the existing client-side rule and the ISO requirement (auto-fill uses a separate path and already supplies a justification).

Added service tests for the invalid-questionId, NO-without-justification, and preserve-on-omit cases (API SoA suite: 58 passing; app SoA suite: 13 passing).

tofikwest:
Thanks @cubic-dev-ai — done in 56b4132c2. In addition to the write-path validation added earlier (`saveAnswer` rejects a questionId not in the document's configuration), `countAnsweredAnswers` is now also scoped to the document's configured question IDs:

```ts
where: {
  documentId,
  isLatestAnswer: true,
  isApplicable: { not: null },
  questionId: { in: validQuestionIds },
}
```

Both callers (manual save and auto-fill) pass the configured IDs, so stale or mismatched answer rows — e.g. an answer for a control later removed from the configuration — can no longer inflate the completion count. Added `soa-storage.spec.ts` with a regression test covering the scoping and the empty-configuration case (API SoA suite now 60 passing).

tofikwest:
@cubic-dev-ai review it

tofikwest:
Thanks @cubic-dev-ai — good catch, fixed in bd7aaa7aa by removing the frontend-only override so both channels read the same source.

The "fully remote → physical-security (7.x) controls are not applicable" rule is already enforced at the data layer: auto-fill persists `isApplicable = false` with the fully-remote justification (`createFullyRemoteResult`), and the field is edit-locked for fully-remote orgs. So the persisted answer already reflects the rule. `resolveSoaDisplay` now just reads the persisted answer (or the in-session autofill result), exactly like the export does — so the on-screen SoA and the PDF always agree.

I chose removing the override over re-applying the rule in the export + answered-count because the latter would duplicate the same rule across four places (display, export, count, generation) and force the export service to look up the org's fully-remote status — the drift risk this review has been guarding against. Enforcing it once at generation and reading the persisted value everywhere keeps a single source of truth. Updated the helper's tests accordingly (app SoA suite: 12 passing).

tofikwest:
@cubic-dev-ai review it

tofikwest:
Thanks @cubic-dev-ai — real bug, fixed in 2cf26d600.

`saveAnswer` now runs all validation (question-in-config, and not-applicable-requires-justification) **before any write**, and the retire-previous + create-new steps are wrapped in a single `db.$transaction`, so a rejected or failed save can never leave a control without a latest answer. Applied the same atomic retire+create to the auto-fill save path (`saveAnswersToDatabase`) so both write paths are consistent. Added a regression test asserting a failed validation does not retire the prior answer (no `update`/`$transaction` calls). API SoA suite: 61 passing.

I also did a full pass over the diff for the same class of issue: those are the only two retire-then-create sites, both now atomic; all validation precedes writes; and no per-org data is read from the shared configuration anywhere.

PR #3397, opened 2026-07-13
fix(soa): apply the fully-remote physical-control rule consistently

tofikwest:
## Summary

Follow-up to #3395, addressing the automated review on the merged SoA change.

A fully remote organization's physical-security (**7.x**) controls are **Not Applicable**. That rule is enforced at generation time (auto-fill persists `isApplicable = false`) and the field is edit-locked — but the on-screen SoA and the PDF export did **not** apply it when no answer was persisted yet. On a freshly created SoA, those controls could render as **Applicable/N/A** and **disagree between screen and export**.

This applies the rule **identically** in both places:
- `resolveSoaDisplay` (screen) forces Not Applicable for fully-remote 7.x controls.
- `exportDocument` applies the same rule (via `checkIfFullyRemote` + `isPhysicalSecurityControl`).

So a fully remote org always sees Not Applicable for physical-security controls **everywhere**, even before auto-fill has persisted it — screen and PDF stay in agreement.

## Changes
- `soa-display.ts`: enforced fully-remote 7.x rule restored (applied consistently, not a display-only override).
- `SOATableRow` / `SOAMobileRow`: pass `isFullyRemote`/`isControl7` through.
- `soa.service.ts` `exportDocument`: apply the same rule so the PDF matches the screen.
- Tests added for both the display helper and the export path.

## On the reviewer's other note (legacy applicability shows N/A)
Documents generated before #3395 don't have a per-answer applicability value, so they show **N/A** until re-generated. This is intentional: the legacy value isn't reliably recoverable from stored data, so surfacing N/A (and regenerating via Auto-fill) is correct rather than backfilling an unreliable value. This is a one-time data-remediation step handled outside this PR — not a code change.

## Testing
- `apps/api`: `npx jest src/soa` — 62 passing (incl. a new export test asserting fully-remote 7.x is forced Not Applicable).
- `apps/app`: `npx vitest run .../statement-of-applicability` — 13 passing (incl. the fully-remote display case).
- Typecheck: no new errors in touched files.

---
## Summary by cubic
Apply the fully-remote physical-security rule consistently across the SoA UI and PDF: for fully-remote orgs, all 7.x controls are forced to Not Applicable and always show the standard remote justification. Fixes screen/export mismatches on new SoAs and prevents stale or cross-org text in those justifications (partially addresses Linear CS-731).

- **Bug Fixes**
  - Enforced the rule in `resolveSoaDisplay`; `SOATableRow`/`SOAMobileRow` now pass `isFullyRemote` and `isControl7`.
  - In `SOAService.exportDocument`, used `checkIfFullyRemote` + `isPhysicalSecurityControl`; when forced, set `isApplicable = false` and used the standard remote rationale for both `answer` and `justification`, overriding any persisted text.
  - Added tests for UI and export, including overriding stale justifications and leaving non‑7.x controls unchanged.

<sup>Written for commit e01fc748668b2516d1d4226c226d23de38e47d10. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3397?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/soa/soa.service.spec.ts
+++ b/apps/api/src/soa/soa.service.spec.ts
@@ -643,6 +643,13 @@ describe('SOAService', () => {
         },
         approver: null,
         answers: [
+          // A stale justification persisted before the org went remote — must
+          // NOT leak into the forced Not Applicable output.
+          {
+            questionId: 'q-phys',
+            answer: 'We maintain physical access controls at our office',
+            isApplicable: true,
+          },
           { questionId: 'q-other', answer: 'Our own 5.1 justification', isApplicable: true },
         ],
       });
@@ -654,9 +661,12 @@ describe('SOAService', () => {
       const other = passedQuestions.find((q) => q.id === 'q-other');
 
       // Fully remote + 7.x → forced Not Applicable with the remote justification,
-      // even though this org has no persisted answer for it.
+      // overriding the stale persisted answer.
       expect(phys?.columnMapping.isApplicable).toBe(false);
       expect(phys?.columnMapping.justification).toBe(
+        'This control is not applicable as our organization operates fully remotely.',
+      );
+      expect(phys?.answer).toBe(
         'This control is not applicable as our organization operates fully remotely.',
       );
       // Non-physical controls are unaffected and use the org's own answer.
--- a/apps/api/src/soa/soa.service.spec.ts
+++ b/apps/api/src/soa/soa.service.spec.ts
@@ -7,6 +7,7 @@ import {
 import { db } from '@db';
 import { SOAService } from './soa.service';
 import { generateSOAExportFile } from './utils/export-generator';
+import { checkIfFullyRemote } from './utils/soa-storage';
 
 jest.mock('@db', () => ({
   db: {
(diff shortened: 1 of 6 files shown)


tofikwest:
Fixed in e01fc7486 — both paths now use the remote rationale unconditionally for forced-remote 7.x controls.

One clarification on the framing: the persisted `SOAAnswer.answer` is per-organization **clean** — the cross-tenant contamination only ever lived in the shared config, which this work stopped reading. So the persisted value here isn't another org's text. **But the fix is still correct:** these controls are edit-locked and forced to Not Applicable, so a persisted justification written *before* the org went remote (their own, but now stale — e.g. "we maintain physical access controls at our office") would contradict the Not Applicable status. Using `FULLY_REMOTE_JUSTIFICATION` unconditionally on both the screen (`resolveSoaDisplay`) and the export keeps the output internally consistent.

Tests updated on both paths to assert a stale persisted justification is overridden by the remote rationale (API SoA: 62 passing; app SoA: 13 passing).

PR #3427, opened 2026-07-16
fix(questionnaire): strip whole-sheet dataValidations before ExcelJS parse

tofikwest:
## Why

api.trycomp.ai went down 2026-07-16 05:39–05:45 UTC (Better Stack alert). Root cause: a customer uploaded `HECVAT_Full.xlsx` to the questionnaire auto-answer export endpoint. The stock HECVAT Full template ships a `<dataValidation>` covering `N1:XFD1048576` + several full-column ranges — **~17.2 billion cells**. ExcelJS registers validations cell-by-cell (`data-validations-xform.js`), so `workbook.xlsx.load()` hard-blocks the Node event loop. ALB health checks (`/v1/health`, 5s timeout) timed out, ECS killed the task, the customer retried, and the replacement died too — three containers were serially killed (05:27, 05:38, 06:02 UTC); the 05:38 overlap took both targets out and caused the visible outage.

The parse never succeeded — the customer never got their export.

## What

- New `stripXlsxDataValidations` (`apps/api/src/utils/sanitize-xlsx.ts`): removes `<dataValidations>` blocks from every worksheet XML inside the xlsx zip before the workbook reaches ExcelJS. Validations hold only dropdown rules and prompts — never cell text — so extraction output is unchanged.
- Wired into `loadWorkbook` in both Excel parse sites: `questionnaire/utils/content-extractor.ts` and `trigger/vector-store/helpers/extract-content-from-file.ts`.
- Safety: best-effort — if there is nothing to strip the original bytes pass through untouched (same reference); if the rewritten archive somehow fails to load, we retry with the original bytes, so worst case equals current behavior. The existing XML fallback path is also still in place.
- ExcelJS ignores `x14` (extLst) validations entirely, so only the standard block needs stripping.

## Verification

- Ran the customer's actual file from S3 through the patched extractor: **hang (killed after 3+ min) → 0.33s, 601,303 chars extracted across all 15 sheets.**
- 28 tests pass across 3 suites, including a new regression test that loads a workbook poisoned with the HECVAT-style full-sheet validation through `extractContentFromFile` (without the fix this test hangs, by design).
- `tsc --noEmit`: only the 23 pre-existing errors already present on main; none in touched files.

## Out of scope (follow-up)

Moving the 4 in-process export parse call sites (`parse/upload`, `parse/upload/token`, JSON export variants) onto the existing Trigger.dev questionnaire tasks, plus a per-file size cap and zip decompression guard — planned as a separate PR since it changes the endpoint contract.

---
## Summary by cubic
Strip Excel data validations from uploaded `.xlsx` files and add a decompression safety check before `ExcelJS` parsing to prevent hangs and zip-bomb crashes. A shared loader keeps questionnaire and vector-store exports fast and stable without changing extracted text.

- **Bug Fixes**
  - Shared loader `loadXlsxWorkbook` uses `stripXlsxDataValidations` and is called by both questionnaire and vector-store paths.
  - Zip-bomb guard `assertXlsxDecompressionWithinLimit` (512 MB uncompressed) runs before any inflation, including ahead of the XML fallback path.
  - Fail-closed: sanitizer throws if it cannot rewrite the archive; no retry with original bytes; questionnaire falls back to the XML extractor so unsanitized bytes never reach `ExcelJS`.
  - Tests added for validation stripping, loader, decompression limit, and a HECVAT regression; customer file now completes in ~0.33s instead of hanging.

<sup>Written for commit d7a3f2dfc191db009274950b4bd3d7cd6ce1f33a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3427?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/questionnaire/utils/content-extractor.spec.ts
+++ b/apps/api/src/questionnaire/utils/content-extractor.spec.ts
@@ -1,4 +1,5 @@
 import { extractContentFromFile } from './content-extractor';
+import AdmZip from 'adm-zip';
 import ExcelJS from 'exceljs';
 import { PDFDocument } from 'pdf-lib';
 import { generateText } from 'ai';
@@ -161,5 +162,44 @@ describe('content-extractor: extractContentFromFile', () => {
     await expect(
       extractContentFromFile(base64, 'application/octet-stream'),
     ).rejects.toThrow('Unsupported file type');
+  });
+
+  // Regression test for the 2026-07-16 api.trycomp.ai outage: HECVAT Full
+  // ships a whole-sheet dataValidation (~17 billion cells) that makes
+  // ExcelJS block the event loop for minutes. Without the sanitizing step
+  // in loadWorkbook this test does not fail — it hangs.
+  it('should extract content from an Excel file with a whole-sheet data validation', async () => {
+    const buffer = await createTestExcelBuffer([
+      {
+        name: 'Questions',
+        rows: [
+          ['Question', 'Response'],
+          ['Will your company sign a BAA?', 'Yes'],
+        ],
+      },
+    ]);
+
+    const zip = new AdmZip(buffer);
+    const entryName = 'xl/worksheets/sheet1.xml';
+    const xml = zip.readAsText(entryName);
+    const validation =
+      '<dataValidations count="1">' +
+      '<dataValidation allowBlank="1" showInputMessage="1" ' +
+      'prompt="Changes cannot be made in this sheet." ' +
+      'sqref="F18:F1048576 N1:XFD1048576 A3:A1048576"/>' +
+      '</dataValidations>';
+    zip.updateFile(
+      entryName,
+      Buffer.from(
+        xml.replace('</worksheet>', `${validation}</worksheet>`),
+        'utf8',
+      ),
+    );
+    const base64 = zip.toBuffer().toString('base64');
+
+    const result = await extractContentFromFile(base64, XLSX_MIME);
+
+    expect(result).toContain('Will your company sign a BAA?');
+    expect(result).toContain('Yes');
   });
 });
--- a/apps/api/src/questionnaire/utils/content-extractor.ts
+++ b/apps/api/src/questionnaire/utils/content-extractor.ts
@@ -5,6 +5,7 @@ import ExcelJS from 'exceljs';
 import AdmZip from 'adm-zip';
 import mammoth from 'mammoth';
 import { PDFDocument } from 'pdf-lib';
+import { stripXlsxDataValidations } from '@/utils/sanitize-xlsx';
 import { PARSING_MODEL, VISION_EXTRACTION_PROMPT } from './constants';
 import { parseQuestionsAndAnswers } from './question-parser';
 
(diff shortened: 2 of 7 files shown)


tofikwest:
Addressed the cubic review in d15660bf7 — dispositions:

**1. P1 — fallback re-parses original bytes (fixed).** Valid: if sanitization changed the archive and the sanitized load failed, retrying the original bytes could reintroduce the hang. The fallback is removed; a sanitized-load failure now propagates (the questionnaire path falls back to the regex-based XML extractor, which can't hang; the vector-store path reports a clean parse error).

**2. P1 — unbounded decompression (fixed, with a caveat).** The sanitizer added no *new* inflation — ExcelJS decompresses the same entries immediately after — but the guard is worth having. `assertXlsxDecompressionWithinLimit` now rejects archives whose declared uncompressed size (read from the zip central directory, nothing inflated) exceeds 512 MB, both inside the shared loader and before the questionnaire XML fallback. Caveat: the check trusts declared entry sizes; a forged central directory could understate them, which no AdmZip-based check can prevent — full streaming-decompression limits belong with the Trigger.dev isolation follow-up.

**3. P2 — regex matches inside CDATA/comments (declined).** Excel never emits unescaped `<dataValidations` literals in worksheet text (cell content is entity-escaped; Excel writes no CDATA in sheet XML), so only a deliberately hand-crafted file could trigger this. With fix 1 in place, the worst case for such a file is a clean parse failure rather than silent corruption or a hang — and an uploader corrupting their own file's extraction is not a meaningful attack. An XML parser here would add cost and its own attack surface for negligible benefit.

**4. P3 — duplicated loadWorkbook (fixed).** Both extraction paths now import the single `loadXlsxWorkbook` from `apps/api/src/utils/load-xlsx.ts`.

Verification unchanged: 33 tests across 4 suites pass, and the customer's actual HECVAT file still extracts 601,303 chars in 0.3s through the reworked path.

tofikwest:
Fixed the remaining P1 in 7a7fdd66d — and it was a fair catch. `stripXlsxDataValidations` swallowed its own failures and returned the original bytes, so an archive that AdmZip couldn't read (but ExcelJS's more lenient unzipper would accept) could reach the parser unsanitized, validations intact. The sanitizer now fails closed: any inspect/rewrite failure throws a descriptive "re-save the file and upload again" error instead of passing unsanitized bytes through. It returns input untouched only when it has positively verified there is nothing to strip.

Trade-off, stated explicitly: a legitimate file that AdmZip cannot read but ExcelJS could have parsed now gets an error instead of being processed. That's the correct direction for this bug class — a rare weird-but-valid file failing loudly beats a crafted file hanging a container — and the error message tells the user how to self-serve (re-save in Excel).

Verification: 33 tests still pass (the not-a-zip fixture now asserts the throw), touched files typecheck clean, and the original customer HECVAT file still extracts 601,303 chars in 0.3s.

PR #3447, opened 2026-07-17
fix(github-integration): distinguish ghas disabled from permission denied in oauth flow

tofikwest:
## Problem
The CodeQL integration incorrectly reports that GitHub Advanced Security (GHAS) is not enabled when using OAuth, even though GHAS is actually active on the repository. This forces customers to manually upload evidence as a workaround and causes confusion about automation support.

## Root cause
The `isGhasEnabled` check in code-scanning.ts conflates two distinct failure modes. It returns false both when `advanced_security.status` is explicitly 'disabled' AND when the `security_and_analysis` block is absent from the API response. Over OAuth, GitHub only returns this block to repository admins due to permission constraints. The check treats a missing block (permission denied) the same as an explicitly disabled status (feature not enabled), returning the 'ghas-required' error in both cases.

## Fix
Updated the logic to distinguish between the two scenarios. When `security_and_analysis` is absent, we now return 'permission-denied' instead of 'ghas-required'. This accurately reflects that we lack permission to read the setting rather than claiming GHAS is disabled. The check still returns 'ghas-required' only when status is positively 'disabled'.

No changes to auth, RBAC, API schemas, or GitHub App integration paths. Behavior for genuinely-disabled repos and the GitHub App connection remains unchanged.

## Explicitly NOT touched
- OAuth token validation or scoping
- GitHub App integration or permissions
- Repository admin checks or RBAC logic
- API schema or response structure

## Verification
✅ Code path for explicitly disabled GHAS still returns 'ghas-required'
✅ Code path for missing security_and_analysis block now returns 'permission-denied'
✅ GitHub App path unaffected (receives full security_and_analysis block)
✅ No schema changes required

Fixes CS-756

---
## Summary by cubic
Fixes false “GHAS not enabled” errors in the OAuth CodeQL check by distinguishing permission issues from GHAS being disabled. Addresses CS-756 and prevents unnecessary manual evidence uploads for private repos.

- **Bug Fixes**
  - Updated `packages/integration-platform/src/manifests/github/checks/code-scanning.ts` to treat a missing `security_and_analysis` block as 'unknown' and return 'permission-denied' on 403 for private repos unless GHAS is explicitly 'disabled'.
  - Only return 'ghas-required' when `advanced_security.status === 'disabled'`; GitHub App behavior is unchanged.
  - Added regression tests in `packages/integration-platform/src/manifests/github/checks/__tests__/code-scanning.test.ts` for OAuth non-admin, GHAS disabled, and GHAS enabled scenarios.

<sup>Written for commit ec42f969ce659735645f4667a775a09883be14ea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3447?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/integration-platform/src/manifests/github/checks/__tests__/code-scanning.test.ts
+++ b/packages/integration-platform/src/manifests/github/checks/__tests__/code-scanning.test.ts
@@ -1,0 +1,135 @@
+import { describe, expect, it } from 'bun:test';
+import type { CheckContext } from '../../../../types';
+import type { GitHubRepo } from '../../types';
+import { codeScanningCheck } from '../code-scanning';
+
+interface RepoConfig {
+  private: boolean;
+  /**
+   * `advanced_security.status`. `undefined` means GitHub omitted the
+   * `security_and_analysis` block entirely — which is what happens over an OAuth
+   * connection whose user lacks repo-admin visibility.
+   */
+  ghas?: 'enabled' | 'disabled';
+  /** Whether the code-scanning default-setup API 403s (permission/GHAS gate). */
+  defaultSetup403?: boolean;
+}
+
+const makeRepo = (fullName: string, config: RepoConfig): GitHubRepo => {
+  const [owner, name] = fullName.split('/');
+  const repo: GitHubRepo = {
+    id: 1,
+    name: name!,
+    full_name: fullName,
+    html_url: `https://github.com/${fullName}`,
+    private: config.private,
+    default_branch: 'main',
+    owner: { login: owner!, type: 'Organization' },
+  };
+  if (config.ghas) {
+    repo.security_and_analysis = { advanced_security: { status: config.ghas } };
+  }
+  return repo;
+};
+
+interface RunResult {
+  passed: Array<{ resourceId: string; title: string }>;
+  failed: Array<{ resourceId: string; title: string }>;
+}
+
+async function runCheck(repos: Record<string, RepoConfig>): Promise<RunResult> {
+  const passed: RunResult['passed'] = [];
+  const failed: RunResult['failed'] = [];
+
+  const ctx = {
+    accessToken: 'tok',
+    credentials: {},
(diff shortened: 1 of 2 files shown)


tofikwest:
re: In code-scanning.ts, the permission-denied remediation is not connection-aware, so OAuth users may be told to change GitHub App permissions they do not control.
→ The permission-denied ctx.fail block (code-scanning.ts:293-311, remediation at :302) is pre-existing and UNCHANGED by this diff — the diff only altered routing logic (isGhasEnabled boolean -> ghasStatus tri-state). It is a human-readable remediation-copy concern, not a logic/correctness defect: pass/fail results, crashes, security, and data are all unaffected. The diff strictly IMPROVES the CS-756 OAuth scenario — those users previously saw a false 'requires GitHub Advanced Security' message and now see the (more accurate) 'cannot access configuration' message; nothing regressed. The finding's proposed fix ('branch on OAuth vs GitHub App') also misdiagnoses the architecture: this integration is OAuth-App-only (index.ts:31-64: type 'oauth2', scopes ['read:org','repo','read:user'], setup instructions 'To create a GitHub OAuth App'), with no GitHub App connection type to branch on. cubic itself flags it non-blocking ('likely safe to merge'). There is a minor real kernel — the pre-existing string says 'GitHub App' / 'Code scanning alerts: Read' (fine-grained GitHub App concepts) which is inaccurate for an OAuth App — but that is a pre-existing, out-of-scope copy nit, not a defect introduced or exposed by this diff.

tofikwest:
@cubic-dev-ai review it

PR #3454, opened 2026-07-20
fix(auth): redirect microsoft sso callback errors to portal instead of api docs

tofikwest:
## Problem
When a new user completes Microsoft SSO during portal onboarding, they are redirected to the Swagger API docs page instead of the portal. This blocks new user signup flow.

## Root cause
The portal initiates Microsoft OAuth via `authClient.signIn.social({callbackURL})` without specifying an `errorCallbackURL`. On the server side, better-auth has no `onAPIError.errorURL` configured. When the OAuth callback encounters any error, better-auth's fallback redirect logic routes to the baseURL (api.trycomp.ai), which has a root redirect to /api/docs, landing the user on Swagger docs.

A June fix addressed one specific cause (email_not_found via upn fallback) but did not fix the error redirect target itself, leaving the issue live for any other callback error.

## Fix
Add `errorCallbackURL` to the portal's Microsoft SSO sign-in call, pointing to the portal origin. This ensures callback errors redirect back to the portal UI instead of the API docs page. No changes to auth logic, RBAC, or server config.

## Explicitly NOT touched
No modifications to better-auth server configuration or global error handling. Auth flow logic and RBAC remain unchanged.

## Verification
Added regression test asserting that portal sign-in specifies a portal-origin errorCallbackURL. Targeted unit tests pass locally ✅

Fixes CS-760

---
## Summary by cubic
Redirect OAuth callback errors back to the portal instead of `api.trycomp.ai` Swagger docs by passing an explicit `errorCallbackURL` for Microsoft (and Google) SSO. Fixes CS-760 and unblocks new user onboarding.

- **Bug Fixes**
  - Added `errorCallbackURL` to `authClient.signIn.social` and ensured success/error URLs are rooted at `portal.trycomp.ai`.
  - Preserves invite and device-auth search params in both success and error redirects.

- **Refactors**
  - Introduced `buildSignInCallbackUrls` used by Microsoft and Google buttons; added `vitest` tests covering invite/device-auth and error paths.

<sup>Written for commit 7920baec27f21571030553bc43b6b63cc779701a. Summary will update on new commits.</sup>

[...]

PR #3459, opened 2026-07-20
fix(people): create invited employees with verified emails so SSO can link

tofikwest:
## Problem

Invited employees frequently could not sign in to the portal with **Google or Microsoft SSO**. The OAuth callback aborted with `account_not_linked` — the dominant Better Auth error in production logs over the last 3 weeks (~70 occurrences, both providers equally).

### Root cause

better-auth refuses to link a trusted OAuth provider to an existing local user whose `emailVerified` is `false` (the `requireLocalEmailVerified` gate, default `true`). Both manual employee-creation paths created users unverified:

- `apps/api/src/people/people-invite.service.ts` (invite flow)
- `apps/app/src/lib/db/employee.ts` (add-people flow)

So: admin invites employee → user row created with `emailVerified: false` → employee clicks "Continue with Google/Microsoft" → better-auth finds the existing user, refuses to link → sign-in fails. (Employees who happened to use email OTP first were unaffected, which is why the failure looked intermittent.)

Related: #3454 (CS-760) fixed *where* the failing user lands after the error; this PR fixes the error itself.

## Fix

1. **Create employee user rows with `emailVerified: true`** in both paths (2-line behavioral change + comments).
2. **Backfill migration** flips existing unverified org members so already-invited employees are unblocked immediately.

### Why this is safe

- **Password auth is disabled.** Every sign-in method (email OTP, magic link, trusted OAuth) proves control of the mailbox before any session is issued — the flag grants nothing to anyone who cannot already receive mail at the address.
- **Precedent:** the employee-sync paths (Google Workspace, Entra, etc.) already create users with `emailVerified: true`. This aligns the manual paths with them.
- **Verified consumers:** the only other `emailVerified` read (`organization-access.service.ts` trycomp.ai auto-approve) uses the *session* user's email behind `HybridAuthGuard` + `organization:update`, so it cannot be influenced by unauthenticated rows.
- Backfill is scoped to users who are org members, and is inert for the same no-password reason.

### Intentionally NOT changed

- `auth.server.ts` / `accountLinking` config: `requireLocalEmailVerified: false` would also fix this, but the option is `@deprecated` in better-auth ("removed on next minor; the gate becomes unconditional") — relying on it would silently re-break on a future upgrade.
- `state_mismatch` (~10 occurrences/3 weeks) is a separate, minor issue (stale tab / double-click), not addressed here.

## Verification

- `cd apps/api && npx jest src/people` — **102/102 pass** (incl. new test asserting invited users are created verified).
- `cd apps/app && npx vitest run src/lib/db/employee.test.ts` — **2/2 pass** (new-user path creates verified; existing-user path never touches the user row).
- API + app typecheck: no new errors (remaining errors are pre-existing on main, verified by side-by-side comparison).

---
## Summary by cubic
Fixes SSO failures for invited employees by creating users with verified emails and upgrading legacy users on re-invite so `better-auth` can link Google/Microsoft accounts, removing `account_not_linked` errors. Also removes unused app-side employee creation code. Addresses Linear CS-760.

- **Bug Fixes**
  - Create users with `emailVerified: true` in the invite flow.
  - On re-invite (employee or admin), auto-verify legacy unverified users; no-op if already verified.
  - Tests cover verified creation, upgrade on re-invite, and no-op for already verified.

- **Migration**
  - Backfill org members with `emailVerified: false` to `true`; safe since password auth is disabled and all sign-in methods prove mailbox control.

<sup>Written for commit 3305c3b8941f01ee19629455541349ff027edd25. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3459?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tofikwest:
Addressed the review findings in 3305c3b89 — disposition of each:

**1. Existing-user invite branch left legacy rows unverified (P2 — valid, fixed).** Both invite paths (employee add + role invitation) now upgrade an existing unverified row via a shared `ensureEmailVerified` helper (no-op when already verified). This covers rows that predate created-verified behavior and were missed by the member-scoped backfill (e.g. their previous org was deleted). Tests added for both paths plus the already-verified no-op.

**2. `employee.ts` creation path reachable without `member:create` (P1 — stale premise, resolved by deletion).** `createEmployeeAction` had **zero callers** anywhere in the app — the live add-people flow goes through the API, where `POST /v1/people/invite`, `POST /v1/people`, and `/bulk` all enforce `member:create`. Rather than leave an unreferenced action that creates users/members with no permission check of its own, the action and its `completeEmployeeCreation` helper are deleted; `lib/db/employee.ts` now exports only `createTrainingVideoEntries` (the one function actually used). Net −217 lines.

**3. `emailVerified` semantics with trusted-provider linking (P1 — not a defect introduced here).** The `OrganizationAccessService.isEmailVerified` concern doesn't apply: that check runs on the *session* user's email behind `HybridAuthGuard`, so it can't be influenced by rows nobody can authenticate as (password auth is disabled; every sign-in method proves mailbox ownership first). The broader observation about trusted-provider linking semantics is a property of the existing better-auth configuration that applies equally to all previously-verified users (every OTP/magic-link user and all sync-created users, which were already created verified) — it is not widened in a new way by this PR and is better handled as its own auth-layer review if desired.

Verification: `npx jest src/people` 105/105 pass; API and app typecheck show zero new errors vs main baselines.

tofikwest:
@cubic-dev-ai  review it

PR #3471, opened 2026-07-21
fix(policies): create draft version on policy regenerate instead of overwriting published

tofikwest:
## Problem

When regenerating a policy, the published version's content is overwritten immediately. This bypasses the approval workflow that should gate any content changes. Regenerating a published policy currently mutates its live content, signedBy metadata, and currentVersionId directly without creating a draft for review.

## Root cause

updatePolicyInDatabase in update-policy-helpers.ts rewrites policy.content, currentVersionId, and signedBy on the live published policy record without changing status or requiring approval. Every other content-change path (createVersion, submitForApproval, acceptChanges) creates a draft version first and only publishes on explicit approval; regenerate was the sole exception.

## Fix

For published or needs_review policies, regenerate now creates a new draft version (incrementing the version number) with the regenerated content, leaving the published version's content, currentVersionId, signedBy, and status unchanged. This mirrors the existing approval workflow: draft versions are created, then approved, then published with the signing flow triggered for all employees.

## Explicitly NOT touched

- Policies not yet published (status: draft)
- The signing flow (runs only on explicit publish, unchanged)
- Approval/rejection workflows
- Version history or deletion logic

## Verification

Added regression test asserting that regenerating a published policy creates a new draft version without modifying the published content or status. Existing policy regenerate and versioning unit tests pass locally ✅

Fixes CS-766

---
## Summary by cubic
Regenerating a policy now creates a draft version for published/needs_review policies and preserves live content and signatures; draft policies are updated in place so the editor shows the new content immediately. Aligns with CS-766 by restoring the approval flow; publishing the draft triggers the signing flow as expected.

- **Bug Fixes**
  - Published/needs_review: create a new draft version; keep published content, `currentVersionId`, `signedBy`, PDFs, and existing versions unchanged.
  - Draft: overwrite the current draft version and sync `content` and `draftContent`; do not append a new version; clear stale PDF references and switch `displayFormat` to `EDITOR` if the draft was a PDF.
  - Stop deleting existing versions and their PDFs (no S3 deletes).
  - Add retry on version creation to avoid unique-key races.
  - Update UI copy: toast says “New draft version created for review” and the dialog explains the new draft flow.
  - Add tests for both published and draft paths, including PDF-to-editor migration.

<sup>Written for commit cadbbe1055db2f89a624bfac26c931ac73b423a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3471?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/trigger/policies/update-policy-helpers.spec.ts
+++ b/apps/api/src/trigger/policies/update-policy-helpers.spec.ts
@@ -246,16 +246,21 @@ describe('updatePolicyInDatabase (draft policy regeneration)', () => {
   beforeEach(() => {
     jest.clearAllMocks();
 
-    // A draft policy: v1 is the current, unpublished, unsigned working version.
+    // A draft policy uploaded as a PDF: displayFormat is 'PDF' and both the
+    // policy and its current version carry a stale pdfUrl (the old document).
     (db.policy.findUnique as jest.Mock).mockResolvedValue({
       id: 'pol_1',
       status: 'draft',
       currentVersionId: 'pv_1',
+      displayFormat: 'PDF',
+      pdfUrl: 'org_1/policies/pol_1/uploaded.pdf',
       content: [
         { type: 'paragraph', content: [{ type: 'text', text: 'Stale draft' }] },
       ],
       signedBy: [],
-      versions: [{ id: 'pv_1', pdfUrl: null, version: 1 }],
+      versions: [
+        { id: 'pv_1', pdfUrl: 'org_1/policies/pol_1/v1.pdf', version: 1 },
+      ],
     });
 
     txPolicyUpdate = jest.fn();
@@ -305,5 +310,21 @@ describe('updatePolicyInDatabase (draft policy regeneration)', () => {
       'Regenerated draft content',
     );
     expect(policyUpdate).not.toHaveProperty('currentVersionId');
+  });
+
+  it('clears stale PDF references and switches to EDITOR display when the draft was uploaded as a PDF', async () => {
+    await updatePolicyInDatabase('pol_1', REGEN_CONTENT, 'mem_regen');
+
+    // Regeneration produces EDITOR content: the policy must switch back to the
+    // editor and drop its stale policy-level PDF, otherwise the page opens on
+    // the PDF tab / export keeps serving the old uploaded document.
+    const policyUpdate = txPolicyUpdate.mock.calls[0][0].data;
+    expect(policyUpdate.displayFormat).toBe('EDITOR');
+    expect(policyUpdate.pdfUrl).toBeNull();
+
+    // The current version's stale PDF (used first by render/export via
+    // currentVersion.pdfUrl ?? policy.pdfUrl) must be cleared too.
+    const versionUpdate = txVersionUpdate.mock.calls[0][0];
+    expect(versionUpdate.data.pdfUrl).toBeNull();
   });
 });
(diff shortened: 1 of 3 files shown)


tofikwest:
@cubic-dev-ai review it

tofikwest:
re: POLICY_VERSION_CREATE_RETRIES duplicates the service-layer retry setting and can drift, so centralize it to a shared source.
→ Maintainability suggestion, not a defect introduced or exposed by this diff. The service value versionCreateRetries=3 is a private readonly instance field (apps/api/src/policies/policies.service.ts:73) that genuinely cannot be imported, and no shared policy constants file exists in apps/api, so the trigger helper cannot reference it directly. The duplicated constant (update-policy-helpers.ts:72) already equals 3 and carries an explicit mirror comment (:69-71). The retry count is best-effort P2002 unique-constraint-race handling; any future divergence would only change retry attempts under a rare collision, never produce incorrect behavior. Cubic itself frames it as 'can drift over time' — a hypothetical future maintenance risk, not a present bug. No correctness impact in the current code.

tofikwest:
@cubic-dev-ai review it

tofikwest:
re: The duplicated helper in apps/app/src/trigger/tasks/onboarding/update-policies-helpers.ts still runs during app Trigger policy regeneration and can overwrite the published policy version / clear signatures for users.
→ App and API are SEPARATE Trigger.dev projects (apps/api/trigger.config.ts proj_zhioyrusqertqgafqgpj vs apps/app/trigger.config.ts proj_lhxjliiqgcdyqbgtucda). tasks.trigger('update-policy') resolves only within the calling process's own project, so it cannot reach the other project's task. Every LIVE regeneration of an existing policy routes to the FIXED API helper: the 'Regenerate policy' button (PolicyHeaderActions.tsx:111 -> usePolicy.ts:98) calls POST /v1/policies/:id/regenerate (policies.controller.ts:382), which runs tasks.trigger('update-policy') at policies.controller.ts:446 inside the API project -> apps/api/src/trigger/policies/update-policy.ts -> fixed update-policy-helpers.ts. Admin regenerate (admin-policies.controller.ts:182) and framework-driven generation (frameworks.service.ts:808) likewise trigger within the API project. The app-side buggy copy is reachable only through the APP project via triggerPolicyUpdates (onboard-organization-helpers.ts:1324), whose callers are (a) onboarding (onboard-organization.ts:108 -> updateOrganizationPolicies -> 1481) operating on freshly-seeded, unsigned draft policies where deleting versions and clearing signedBy is the correct initial behavior, and (b) generate-full-policies.ts:43, triggered only by regenerateFullPoliciesAction (regenerate-full-policies.ts:24) which has ZERO callers (dead code). The visible 'Regenerate all' button (FullPolicyHeaderActions.tsx:35 -> usePolicyActions.ts:8) posts to /v1/policies/regenerate-all, which does NOT exist in the API (no such route in policies.controller.ts), and FullPolicyHeaderActions is never mounted (only its own .test.tsx imports it). Thus no user regenerating a published/signed policy reaches the app-side copy, and this diff did not change app-side reachability. The duplication is a real maintenance smell (consolidating to a shared helper is good hygiene), but the specific defect claimed is not reachable, so it is not a genuine user-facing defect introduced or exposed by this diff. Residual uncertainty: 'Regenerate all' is clearly intended as a live feature, so if that endpoint/wiring is ever restored it would hit the buggy copy — but that is a separate, pre-existing, currently-dead path, not what this PR fixes or what the finding describes.

tofikwest:
@cubic-dev-ai review it

PR #3495, opened 2026-07-23
fix(device-registration): set source to device_agent when agent adopts intune device

tofikwest:
## Problem
A device migrated from Intune integration to Comp AI endpoint agent shows as "missing" in the People tab (Overview) but displays correctly in the Device tab. The customer reported the device was initially imported via Intune, Intune was disconnected, the endpoint agent was installed and confirmed working, but the following morning the device appeared as missing in the People tab only.

## Root cause
When a device migrates from Intune to the endpoint agent (same serial + member), the agent adoption flow in `registerWithSerial` (device-registration.helpers.ts) and `checkIn` (device-agent-auth.service.ts) updates the existing Device row but never sets the `source` field. The row retains `source='integration'` while carrying live agent compliance data. The People tab rollup query filters to only `source='device_agent'` devices (compute-device-status-map.ts:41), so the device is excluded and the member renders as "not-installed" → "Missing". The Device tab lists all devices regardless of source, so it remains visible there. This asymmetry matches the customer's report.

## Fix
Set `source: 'device_agent'` in the agent's adopt and check-in update paths. The change is localized to two code paths (registerWithSerial and checkIn) and backward-compatible: agent rows are already sourced correctly; the Intune sync already refuses to overwrite non-integration serials, so no race condition. No schema change required.

## Explicitly NOT touched
- Offboarding date handling (the customer's mention of an incorrect offboardDate is a separate minor issue and does not affect device display logic, which gates on deactivated/isActive flags)
- The "unable to clear offboardDate" UI issue in EmployeeDetails.tsx (separate bug)
- Device deactivation or member lifecycle logic

## Verification
- Added regression test asserting that when a device transitions from integration source to agent source via checkIn, the source field is updated to 'device_agent' and the device appears in the People-tab rollup query result
- Existing device-registration and device-agent-auth unit tests pass locally ✅

Fixes CS-770

---
## Summary by cubic
Fixes devices migrating from Intune to our endpoint agent showing as Missing in People by setting device `source` to `agent` during adoption and check-in. Addresses CS-770 so devices show consistently in both People and Devices.

- **Bug Fixes**
  - Set `source: 'agent'` in `registerWithSerial` and `checkIn` when the agent adopts an integration-imported device.
  - Added tests in `device-agent-auth.service.spec.ts` and `device-registration.helpers.spec.ts` to assert the source update.

<sup>Written for commit 4d39ef79df69cfc42fd229413a6300979b4f2b5f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3495?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tofikwest:
re: In apps/api/src/device-agent/device-agent-auth.service.ts, the CS-770 migration guard is not race-safe: an in-flight Intune sync can read the old source and overwrite the new assignment, flipping migrated devices back to integration; make the source update/check atomic.
→ The flagged file has no racy read-check-write on source. checkIn (device-agent-auth.service.ts:221-236) issues one db.device.update that sets source:'agent' UNCONDITIONALLY — the written value never depends on the device it read (findFirst at :159), so its write is a single atomic UPDATE. Same in device-registration.helpers.ts buildUpdateData (:22), unconditional source:'agent'. The described 'migration guard [that] is not race-safe' does not exist in this file. The only source read-check-write is the pre-existing, untouched generic-device-sync.service.ts:142 (`source !== 'integration'` -> skip), a deliberately deferential guard that leaves any non-integration device alone. device.prisma:32 shows `source @default(agent)`, so agent-created devices were always 'agent' and always covered by that same sync guard — the theoretical sync-vs-agent interleaving predates this diff and applied to every agent device; this PR only broadens the already-protected 'agent' class to adopted rows, introducing/exposing no new race. The suggested remedy ('conditional write tied to the expected current source') would actively break CS-770: the check-in must assert ownership regardless of prior source — that IS the fix — so there is no 'expected current source' to gate on. Finally the model is intentionally eventually-consistent: the check-in re-asserts 'agent' unconditionally ~hourly while syncs run far less often and defer via the guard, so any sub-millisecond interleaving (no awaits between the sync's serial-matched read at :130 and update at :207) self-heals on the next check-in. Not a defect introduced or exposed by this diff.

tofikwest:
@cubic-dev-ai review it

PR #3501, opened 2026-07-24
fix(training): remove rbac gate from mark-complete endpoint

tofikwest:
## Problem

A user with a custom role lacking portal:update permission is unable to mark training videos complete, receiving a 403 error and seeing "failed to mark video as completed" in the UI. The user can accept policies and complete device setup successfully, but gets blocked at the training endpoint.

## Root cause

The POST /v1/training/completions/:videoId/complete endpoint is gated by @RequirePermission('portal','update') in training.controller.ts. This creates an inconsistent permission model: policy acceptance uses session + member-ownership only (no RBAC check), while training completion enforces a portal:update requirement that custom roles may lack. Existing custom roles created before permission enforcement was added still lack this permission.

## Fix

Remove the @RequirePermission('portal','update') guard from the mark-complete endpoint. Training completion is a user action on their own record (like policy acceptance), not a privileged administrative operation, so it should authenticate by session + member-ownership only, not RBAC.

## Explicitly NOT touched

Policy enforcement at role creation time (from PR #3455) remains in place. Other training endpoints and their permission gates are unchanged. Portal write permissions on other operations are unaffected.

## Verification

Added regression test asserting that a user with a custom role lacking portal:update can successfully mark a training video complete. Existing training controller unit tests pass locally ✅.

Fixes CS-774

---
## Summary by cubic
Fixes CS-774 by letting employees on custom roles mark training videos complete without `portal:update`. Completion now uses a portal self-service route and triggers completion emails when eligible.

- **Bug Fixes**
  - Added POST /api/portal/complete-training (session + org membership, no RBAC) with video ID validation, HIPAA gating for `hipaa-sat-1`, atomic upsert, no re-stamps, 400s on invalid JSON/IDs, and production logging of non-2xx email responses via `console.error`. Tests cover 401/403, invalid JSON/IDs, HIPAA gating, concurrency, re-stamps, and prod logging on email failures.
  - Updated `use-training-completions` to call the portal route, include `organizationId`, and read the `{ data }` response.

- **New Features**
  - Added POST /v1/training/send-hipaa-completion-email (requires `training:update`); the portal route triggers HIPAA or general completion emails when all videos are complete. Controller tests verify the permission gate.
  - Updated Vitest aliases for `@db/server` and `@db`.

<sup>Written for commit 70cad54d678716271ad79aa458815f893c888634. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3501?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tofikwest:
re: Repeated completion requests trigger duplicate certificate emails (email not idempotent)
→ The non-idempotency is real in isolation (training-email.service.ts has no dedup; EmployeeTrainingVideoCompletion model at packages/db/prisma/schema/auth.prisma:43-53 has no emailSent flag; route.ts calls sendCompletionEmailIfComplete on every POST, even the already-completed no-restamp branch). BUT this is NOT introduced or exposed by this diff: the replaced service method markVideoComplete (training.service.ts:86-110) already sent the completion/HIPAA email unconditionally on every invocation with identical no-guard behavior, and the UI (HipaaTrainingAccordionItem hides the button once isCompleted; general videos likewise) does not re-trigger. The self-service trigger surface is equivalent to before, so this is a pre-existing latent behavior, not a regression attributable to this PR. Worth fixing generally, but not a defect this diff creates.

tofikwest:
@cubic-dev-ai review it

tofikwest:
re: Route bypasses TrainingController's portal:update gate; custom-role users still 403 — route it through the same authorization gate.
→ This inverts the entire purpose of the PR. CS-774's fix is precisely to STOP routing portal self-service through the RBAC-gated NestJS endpoint. The hook use-training-completions.ts:47 now POSTs to /api/portal/complete-training instead of /v1/training/completions/:videoId/complete, so the RBAC gate is never hit for this flow. The new route authorizes on session + active org membership (route.ts:36-69), identically to the existing accept-policies/route.ts and mark-policy-completed/route.ts it mirrors. Re-adding the portal:update gate would re-introduce the exact 403 bug the PR fixes.

re: Org scoping forgeable via request body; derive org from session.portalOrganizationId instead of client-supplied organizationId.
→ The suggested fix references session.portalOrganizationId, a field that does not exist anywhere in the codebase (grep returned no matches); the portal is inherently multi-org, selecting org via the URL orgId param (hook line 36-37), so there is no single 'active portal org' pinned in the session. Authorization is correct: member is looked up by {userId: session.user.id, organizationId, deactivated:false} (route.ts:59-65) and the completion record is written with memberId: member.id (route.ts:95), so a user can only mark THEIR OWN training complete in an org they are an active member of. No cross-tenant or cross-user forgery is possible. The mirrored routes (accept-policies, mark-policy-completed) derive the member from the session the same way and do not use any portal-org session field. Completing your own training in an org you legitimately belong to is not privilege escalation.

tofikwest:
@cubic-dev-ai review it

tofikwest:
@cubic-dev-ai review it

PR #3514, opened 2026-07-28
feat(policies): add bulk upload for policy migration

tofikwest:
## Problem
Users migrating policies from another platform need a way to bulk upload pre-existing policies instead of creating them one-by-one in the policies tab.

## Root cause
The policies tab lacked bulk import functionality, forcing manual creation of each policy during platform migrations.

## Fix
Added bulk upload feature to the policies tab that accepts a batch of policies and creates them in the system. The implementation includes:
- Upload handler to accept policy files/data
- Validation of incoming policies against the schema
- Batch creation in the database
- User feedback on success/failure per policy

## Explicitly NOT touched
- Individual policy creation flow (existing UI and validation remain unchanged)
- Policy editing or deletion
- Export functionality

## Verification
- Added regression test asserting that multiple policies can be bulk uploaded and appear in the policies list with correct attributes ✅
- Existing policy unit tests pass locally ✅

---
## Summary by cubic
Adds bulk upload to the Policies tab so teams can import multiple policy PDFs at once during migrations. Each PDF creates a draft policy and attaches the document, with per-file results, retries that reuse drafts, and resilience to thrown network errors.

- **New Features**
  - Adds a “Bulk upload” button (requires `policy:create` and `policy:update`) that opens `BulkUploadPoliciesSheet`.
  - Accepts PDFs only, up to 25 files (100MB each); de-dupes by name+size and shows toasts for rejections.
  - Uploads directly to `/v1/policies` then `/v1/policies/:id/pdf` with bounded concurrency and lazy base64 reads to limit memory.
  - Shows created/failed counts and refreshes the list on success; on partial success, keeps only failed files with inline errors, reuses created drafts on retry, and blocks closing while uploads run.
  - Tests cover the button/sheet and permission gating, bulk upload orchestration (create/attach, failure paths, draft reuse, concurrency), and blocking close mid-upload.

- **Bug Fixes**
  - Isolates a thrown `post()` during create/attach so one bad request doesn’t reject the whole batch; records a failed row and preserves any created draft id for retry. Regression test added.

<sup>Written for commit ca7ac76458129831bf1997fe3b1b3820d8a10589. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3514?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tofikwest:
re: #2 — route trusts client constraints; no server-side file-size / aggregate / real-PDF-type validation
→ Largely inaccurate or mitigated. Aggregate batch size IS enforced server-side (`MAX_FILES=25`, route.ts:33-38), so that claim is wrong. Byte size is effectively bounded upstream — Vercel's ~4.5MB function-body cap (see #1) plus NestJS `express.json({limit:'150mb'})` (main.ts:110-116) — so 'oversized content reaches storage' is contradicted by the very cap that makes #1 a bug; there is no unbounded-buffer path on the real deployment. Magic-byte/MIME validation is indeed absent, but (i) it is the SAME shared endpoint the pre-existing PdfViewer single-upload already uses (controller:567-648) — not introduced by this diff — and (ii) practical stored-XSS is neutralized because all download endpoints force `ResponseContentType:'application/pdf'` (controller:549-550, 958-959), and the endpoint only stores bytes in S3 (no server-side parsing). The endpoint also already rejects non-base64 / empty data (controller:634-641). Remaining item is low-value defense-in-depth hardening, not a defect this diff introduces.

tofikwest:
@cubic-dev-ai review it

tofikwest:
@cubic-dev-ai review it

tofikwest:
re: Uncaught throws from injected post() can reject the entire batch, hiding per-file outcomes for remaining uploads.
→ The only injected post is useApi().post (use-api.ts:26-29) → api.post → ApiClient.call (api-client.ts:42-77). call() wraps its whole body in try/catch and ALWAYS returns a structured ApiResponse; even transport/network failures resolve to { status: 0, error: 'Network error' } (lines 71-76). JSON.stringify runs on plain string payloads before call() (no circular refs/BigInt), so it can't throw either. Thus post() never rejects with the real client, and the 'one transport exception aborts the batch' scenario cannot occur. Network errors already propagate per-file (createRes.error/attachRes.error truthy → that file fails, others continue), honoring the per-file-isolation invariant. Purely a defensive-coding suggestion, not a defect reachable in this codebase.

tofikwest:
@cubic-dev-ai review it

tofikwest:
@cubic-dev-ai review it

tofikwest:
re: policies-bulk-upload.ts: post() throwing short-circuits Promise.all and aborts the batch; wrap each file's create/attach in try/catch.
→ The injected post is apiClient.post -> ApiClient.call, which wraps fetch, response.text(), and JSON.parse in try/catch and returns {error, status:0} on any throw (apps/app/src/lib/api-client.ts:42-77) — it never rejects. In uploadOnePolicy the only throwable op, readFileAsBase64, is already try/caught (policies-bulk-upload.ts:86-90); both post calls are branched on the returned .error field, not exceptions (:103, :119). JSON.stringify runs outside call's try, but the bodies are plain strings and [] (:97-100, :113-117) with no circular refs/BigInt, so it cannot throw. Therefore uploadOnePolicy never rejects, worker never rejects, and Promise.all (:188) never aborts. The described failure scenario has no execution path in this checkout; the one genuinely throwable case is already handled.

tofikwest:
@cubic-dev-ai review it

PR #3515, opened 2026-07-28
fix(cloud-security): add missing logGroupName to CreateLogGroup remediation

tofikwest:
## Problem

Auto-remediation for the CloudTrail/CloudWatch Logs compliance finding fails silently. The CreateLogGroup API call rejects the request because no log group name is supplied, leaving the resource non-compliant. Customers who fall back to manual remediation steps also report difficulty completing them independently.

## Root cause

The CreateLogGroupCommand in the auto-remediation flow does not enforce or supply the required logGroupName parameter. Unlike PutMetricFilter (which has both a required-params check and a deterministic name injector), CreateLogGroup lacks these guards. The parameter reaches the AWS API as undefined, causing rejection.

## Fix

Add logGroupName to REQUIRED_PARAMS in aws-command-executor.ts alongside an injector that defaults the name to aws-cloudtrail-logs, mirroring the pattern already established for PutMetricFilter. This ensures the CreateLogGroup call always includes a valid name without manual configuration.

## Explicitly NOT touched

Manual remediation step clarity (addressed separately if needed). Parser or orchestration logic outside cloud-security scope. Any other compliance remediations.

## Verification

Added regression test asserting that CreateLogGroup remediation includes a valid logGroupName parameter and passes validation. Unit tests for aws-command-executor and the injector logic pass locally ✅

Fixes CS-787

---
## Summary by cubic
Fixes CloudTrail → CloudWatch Logs auto-remediation (CS-787). We now validate that `CreateLogGroup` includes `logGroupName` and only backfill a name derived from the trail’s `CloudWatchLogsLogGroupArn` to avoid orphan groups.

- **Bug Fixes**
  - Add `CreateLogGroupCommand: ['logGroupName']` to `REQUIRED_PARAMS`; clear fail-fast error prevents opaque AWS failures across all log-group remediations.
  - Backfill only when a CloudTrail step has a parseable `CloudWatchLogsLogGroupArn`; derive the name, preserve existing names; applied to fix and rollback; no default when no ARN (prevents false success).
  - Scope to CloudTrail plans (`UpdateTrail`/`CreateTrail`); no injection for Step Functions, SSM Session Manager, or Transfer Family. Tests cover missing/present, derived, preserved, rollback, non-CloudTrail, and no-ARN cases.

<sup>Written for commit 6edf348894c43e2e98a266514308571230caa92a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3515?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tofikwest:
@cubic-dev-ai review it

tofikwest:
@cubic-dev-ai review it

tofikwest:
re: Finding 1: missing-ARN fallback leaves UpdateTrailCommand without CloudWatchLogsLogGroupArn, so CloudTrail→CloudWatch integration stays incomplete after creating aws-cloudtrail-logs; derive and inject the log group ARN before emitting the update step.
→ Kernel of truth exists in a narrow edge, but not an actionable defect in this diff. (1) INFEASIBLE remedy: normalizeFixPlan is pure with context={resourceId} only (plan-normalizer.ts:64-66); for this finding resourceId is the checkId 'cloudwatch-no-cloudtrail-integration' (cloudwatch.adapter.ts:359), not an ARN — so no accountId/region. A log-group ARN is arn:aws:logs:{region}:{accountId}:log-group:{name}:*; the diff only PARSES an existing ARN (logGroupNameFromArn). Constructing one would require fabricating account/region, producing an invalid or wrong-account ARN (worse than omitting). (2) NOT introduced by the diff: UpdateTrail-missing-ARN is a pre-existing AI-plan-completeness issue the diff doesn't touch; before the diff this same input hard-failed at CreateLogGroup anyway. (3) Doesn't occur on the realistic path: the finding's remediation text (cloudwatch.adapter.ts:149-150) explicitly instructs UpdateTrail with CloudWatchLogsLogGroupArn, so it's present and inferCloudTrailLogGroupName derives the matching name — coherent. The updateTrailStepMissingArn test only exercises name-defaulting in isolation, not a claim of full-plan correctness. Manifests only in a quadruple-edge (AI omits logGroupName AND ARN AND no sibling ARN AND UpdateTrail no-ops); completing trail linkage is outside the normalizer's documented backfill scope.

re: Finding 2: conditional injector lacks a hard guard when logGroupName absent, so malformed CreateLogGroupCommand steps can still be produced and fail later; fail fast on missing logGroupName and block plan generation for invalid inputs.
→ Misunderstands the normalizer's role and contradicts the codebase pattern. normalizeFixPlan is documented pure/idempotent/non-throwing (plan-normalizer.ts:68-89) and the sibling SLR backfill returns the step unchanged when it cannot infer (plan-normalizer.ts:150-164) — never throws. For the targeted CloudTrail case, backfillLogGroupName ALWAYS resolves a non-empty name (inferred ARN name or default 'aws-cloudtrail-logs', :226-240), so NO malformed CreateLogGroup step survives — the premise is false for the scenario the diff targets. The only steps left with an empty logGroupName are non-CloudTrail plans (SSM, Step Functions, Transfer), which the diff deliberately and correctly leaves alone (:211-216) — 'blocking plan generation' by throwing would break those legitimate, out-of-scope plans. The actual fail-fast layer is validatePlanSteps in aws-command-executor.ts:588-691 via REQUIRED_PARAMS; CreateLogGroupCommand is intentionally not listed there (:161-169) precisely because the backfill handles the CloudTrail case. Adding a throw in the normalizer is the wrong location and wrong mechanism.

tofikwest:
@cubic-dev-ai review it

PR #3532, opened 2026-07-29
fix(people): show device status matching devices tab by filtering stale agent rows

tofikwest:
## Problem

Customer reports that the People page shows devices as "Missing" even though those same devices appear compliant in the Devices tab after a fresh agent install and successful check-in. The mismatch is consistent across multiple devices in their macOS fleet.

## Root cause

Two code paths query device status independently:

1. Employee Device tab queries the DB directly and returns the newest installed agent (getMemberDevice in [orgId]/people/[employeeId]/page.tsx)
2. People page column calls /api/people/agent-devices which rolls up status from all agent rows per member, collapsing stale/non-compliant/not-installed into a single "Missing" badge via compute-device-status-map.ts and MemberRow.tsx:222

When a fresh agent installs and checks in compliant, an orphaned stale agent row (from prior hostname-keyed or fallback-serial registration paths in device-registration.helpers.ts) remains in the DB marked isCompliant=false. The People page aggregation sees this stale row and pins the member to "Missing" forever, even though the current agent is healthy. Additionally, TeamMembersClient.tsx:95-97 silently discards useAgentDevices errors while the Device tab surfaces them, masking failures in the agent-devices route.

## Fix

Filter stale agent rows before status aggregation in the /api/people/agent-devices route. Include only agents with recent activity (within the configured stale threshold) and compliant status. This ensures the People page reflects the same current device state as the Device tab and ignores abandoned agent registrations from prior installs.

## Explicitly NOT touched

- Device tab query logic or getMemberDevice
- Device registration path selection or hostname fallback logic
- Agent check-in or compliance flagging logic
- Error handling in TeamMembersClient (separate concern)

## Verification

- Added regression test asserting that People page device column filters out stale agent rows and shows compliant status when a fresh agent checks in after prior install ✅
- Existing unit tests for compute-device-status-map and /api/people/agent-devices pass locally ✅

Fixes CS-791

---
## Summary by cubic
Fixes People page showing "Missing" while Devices tab shows compliant by deduping agent device rows to the newest per machine and retiring duplicate/ghost rows during registration. Aligns the People roll-up with the Devices tab (Linear CS-791).

- **Bug Fixes**
  - Deduplicate by member+hostname and pick the newest `installedAt` row in People status (`compute-device-status-map`), so older stale rows no longer override the current device.
  - Ignore integration-imported rows when computing People status; only true agent rows affect compliance.
  - Registration now adopts existing rows when serial reads fail or use fallback serials, and removes superseded duplicates to keep one active row per machine.
  - Added focused tests for CS-791 to cover dedupe and registration cleanup paths.

<sup>Written for commit b751d38eb2f9b761f28a6282f1f5d01ca436efc4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/device-agent/device-agent-auth.service.spec.ts
+++ b/apps/api/src/device-agent/device-agent-auth.service.spec.ts
@@ -19,6 +19,7 @@ jest.mock('@db', () => ({
       create: jest.fn(),
       update: jest.fn(),
       findMany: jest.fn(),
+      deleteMany: jest.fn(),
     },
     session: {
       delete: jest.fn(),
--- a/apps/api/src/device-agent/device-registration.helpers.spec.ts
+++ b/apps/api/src/device-agent/device-registration.helpers.spec.ts
@@ -5,6 +5,7 @@ jest.mock('@db', () => ({
       findFirst: jest.fn(),
       update: jest.fn(),
       create: jest.fn(),
+      deleteMany: jest.fn(),
     },
   },
 }));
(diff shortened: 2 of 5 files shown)


tofikwest:
@cubic-dev-ai review it

PR #3535, opened 2026-07-29
fix(device-agent): correct macos password policy check to parse modern account policy

tofikwest:
## Problem

Device Agent incorrectly reports that password policy checks pass even when a user's password is 4 characters, below the required minimum of 8. A customer reported this on Apple Silicon macOS.

## Root cause

The password policy check in `packages/device-agent/src/checks/macos/password-policy.ts` uses three regex patterns to parse the minimum password length from `pwpolicy getaccountpolicies` output. However, these patterns do not match macOS's actual account policy representation. On current macOS versions the format is `policyAttributePassword matches '.{4,}+'` but the agent was looking for `policyAttributePasswordMinimumLength=`, `minChars[=:]`, or `minimumLength[=:>\s]` all of which return null. This causes the check to fall through to deprecated global policy values or loose grepping of system_profiler output, allowing a 4-character password to pass validation.

## Fix

Parsed the correct modern account policy format `policyAttributePassword matches '.{N,}+'` from `pwpolicy getaccountpolicies` output and treat it as authoritative. When both account-level and global policies exist, take the conservative minimum instead of the maximum, ensuring the stricter requirement is enforced.

## Explicitly NOT touched

No changes to other check files or remediation logic. The fix is localized to the password policy check parsing.

## Verification

Added regression test asserting that a 4-character password correctly fails policy validation when the account policy requires a minimum of 8 characters. Unit tests for the password policy check pass locally ✅

Fixes CS-795

---
## Summary by cubic
Fixes the macOS password policy check in `packages/device-agent` by parsing modern account policies and ignoring the deprecated global store, so 4-character passwords no longer pass. Updates remediation to guidance-only and docs to match current macOS behavior. Addresses Linear CS-795.

- **Bug Fixes**
  - Parse `policyAttributePassword matches '.{N,}+'` from `pwpolicy getaccountpolicies` and use the largest required length.
  - Read MDM Passcode `minLength` from `system_profiler` and combine; if none exist, the check fails.
  - Stop consulting `pwpolicy -getglobalpolicy` (deprecated and not enforced).
  - Added regression tests for checks/remediation and updated docs/instructions.

- **Migration**
  - Managed Macs: enforce via an MDM Passcode payload with `minLength >= 8`.
  - Unmanaged Macs: set global account policies with `sudo pwpolicy -setaccountpolicies <file>`; keep the built-in FileVault policy and verify with `pwpolicy getaccountpolicies`.
  - Do not use `pwpolicy -setglobalpolicy "minChars=8"`; it’s deprecated and not enforced.

<sup>Written for commit 12d771f24b7f1b734e26eea1832b29c9b2785a44. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3535?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/device-agent/src/checks/macos/password-policy.spec.ts
+++ b/packages/device-agent/src/checks/macos/password-policy.spec.ts
@@ -1,0 +1,123 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const execSyncMock = vi.fn();
+
+vi.mock('node:child_process', () => ({
+  execSync: (command: string) => execSyncMock(command),
+}));
+
+import { MacOSPasswordPolicyCheck } from './password-policy';
+
+/**
+ * Captured `pwpolicy getaccountpolicies` output of a stock, unmanaged Mac (macOS 26.2),
+ * with the non-English `policyContentDescription` localizations trimmed.
+ */
+const ACCOUNT_POLICIES_MIN_4 = `Getting global account policies
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>policyCategoryPasswordContent</key>
+	<array>
+		<dict>
+			<key>policyContent</key>
+			<string>policyAttributePassword matches '.{4,}+'</string>
+			<key>policyContentDescription</key>
+			<dict>
+				<key>en</key>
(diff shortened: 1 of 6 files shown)


tofikwest:
@cubic-dev-ai review it

PR #3574, opened 2026-09-22
fix(evidence-forms): store uploads under the form's own file field

0xtechdean:
## What does this PR do?

`uploadSubmission` wrote every uploaded file to a hardcoded `evidenceFile` key, regardless of form type. The submission detail view renders each field by its **declared** key (`submission.data[field.key]`), so for any form whose file field is named something else, the real field stayed empty — the submission rendered `—` against every field, including the one its own submission schema marks **required**.

The file wasn't lost: `CompanySubmissionDetailPageClient` has a fallback that appends a generic **Uploaded Evidence** row when `data.evidenceFile` is present and the form declares no field by that name. That row is why this looks superficially fine while the form's actual fields are blank.

Only two forms actually declare a field called `evidenceFile` — `whistleblower-report` and `tabletop-exercise`. Every other form with a file field was storing under a key nothing reads:

| Form | Declared file field | Written as (before) |
|---|---|---|
| `penetration-test` | `pentestReport` *(required)* | `evidenceFile` |
| `network-diagram` | `diagramFile` | `evidenceFile` |
| `rbac-matrix` | `matrixFile` | `evidenceFile` |

So an uploaded penetration test report displayed with no Test date, Vendor / Testing Firm, Summary of Findings **or** Pentest Report against it — just a bare file link. For `network-diagram` the stored row doesn't satisfy its own schema, which requires either `diagramUrl` or `diagramFile`.

### The fix

Resolve the target key from the form definition rather than hardcoding it, preferring a `required` file field when a form declares more than one. Forms that declare no file field keep the generic key, so their behaviour is unchanged.

The generic fallback row is deliberately left in place — it still serves forms with no file field, and existing rows already written under `evidenceFile`.

### Note on existing data

This fixes new uploads only. Submissions already created through this path still hold the file under `evidenceFile` and will keep rendering through the fallback row. They can be migrated by copying `data->'evidenceFile'` to the form's declared key; I've not included a migration here since the right call on back-filling is yours.

## Visual Demo

N/A — the change is to which key the submission payload is written under. The behavioural difference is covered by tests: reverting the resolver to the old hardcoded key fails 5 of them.

- [x] I have self-reviewed the code.
- [x] I have updated the developer docs in /docs if this PR makes changes that would require a documentation change. N/A — no documented behaviour changes; the form definitions are already the source of truth.
- [x] I have added tests: `resolveUploadFileFieldKey` unit tests (including an invariant over *every* registered form type) plus `uploadSubmission` service tests asserting the file lands on the declared field and that the resulting submission passes its own schema. `npx jest src/evidence-forms` → 36 passed, 4 suites.

---
## Summary by cubic
Fixes `uploadSubmission` storing every uploaded file under a hardcoded `evidenceFile` key. Only two forms declare a field named `evidenceFile`; `penetration-test`, `network-diagram`, and `rbac-matrix` each declare their own file field, so uploads left the form's real field empty and the submission rendered "—" against it. The target key is now resolved from the form definition, preferring a required file field when several exist. Forms that declare no file field keep the generic key, so their behavior is unchanged. This fixes new uploads only; existing submissions still hold the file under `evidenceFile` and render via the existing fallback row.

<sup>Written for commit 869606cc14b2ab0debe03a3ed3b20cede6e60258. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3574?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


0xtechdean:
Closing — opened from a fork branch based on our own main, so it carried 38 unrelated commits from our fork rather than just the fix. Reopening from a branch based directly on upstream/main with only the single commit.

PR #3575, opened 2026-09-22
fix(evidence-forms): store uploads under the form's own file field

0xtechdean:
## What does this PR do?

`uploadSubmission` wrote every uploaded file to a hardcoded `evidenceFile` key, regardless of form type. The submission detail view renders each field by its **declared** key (`submission.data[field.key]`), so for any form whose file field is named something else, the real field stayed empty — the submission rendered `—` against every field, including the one its own submission schema marks **required**.

The file wasn't lost: `CompanySubmissionDetailPageClient` has a fallback that appends a generic **Uploaded Evidence** row when `data.evidenceFile` is present and the form declares no field by that name. That row is why this looks superficially fine while the form's actual fields are blank.

Only two forms actually declare a field called `evidenceFile` — `whistleblower-report` and `tabletop-exercise`. Every other form with a file field was storing under a key nothing reads:

| Form | Declared file field | Written as (before) |
|---|---|---|
| `penetration-test` | `pentestReport` *(required)* | `evidenceFile` |
| `network-diagram` | `diagramFile` | `evidenceFile` |
| `rbac-matrix` | `matrixFile` | `evidenceFile` |

So an uploaded penetration test report displayed with no Test date, Vendor / Testing Firm, Summary of Findings **or** Pentest Report against it — just a bare file link. For `network-diagram` the stored row doesn't satisfy its own schema, which requires either `diagramUrl` or `diagramFile`.

### The fix

Resolve the target key from the form definition rather than hardcoding it, preferring a `required` file field when a form declares more than one. Forms that declare no file field keep the generic key, so their behaviour is unchanged.

The generic fallback row is deliberately left in place — it still serves forms with no file field, and existing rows already written under `evidenceFile`.

### Existing data

Included as a backfill migration (`20260922120000_backfill_evidence_upload_file_field`), following the existing `backfill_*` convention. For the three affected form types it copies the file onto the declared key, then drops the generic key so the file renders **once** under its proper field rather than twice — the fallback row would otherwise still fire, since those forms declare no field named `evidenceFile`.

It is deliberately conservative: it only writes where the declared key is absent, and only removes `evidenceFile` after confirming the copy matches, so it is idempotent and safe to re-run. `whistleblower-report` and `tabletop-exercise` (which genuinely declare `evidenceFile`) and the form types with no file field are untouched.

Verified against a real database inside a rolled-back transaction: a simulated pre-fix row converts correctly, the file object is preserved intact, and a second run is a no-op (`UPDATE 0`).

## Visual Demo

N/A — the change is to which key the submission payload is written under. The behavioural difference is covered by tests: reverting the resolver to the old hardcoded key fails 5 of them.

- [x] I have self-reviewed the code.
- [x] I have updated the developer docs in /docs if this PR makes changes that would require a documentation change. N/A — no documented behaviour changes; the form definitions are already the source of truth.
- [x] I have added tests: `resolveUploadFileFieldKey` unit tests (including an invariant over *every* registered form type) plus `uploadSubmission` service tests asserting the file lands on the declared field and that the resulting submission passes its own schema. `npx jest src/evidence-forms` → 36 passed, 4 suites.


0xtechdean:
Good catch on the enum labels — that was a real defect and it would have aborted the whole migration before touching a row, not silently skipped.

Root cause: `EvidenceFormType` is declared with `@map`, so the Prisma client names are underscored but the labels stored in Postgres are hyphenated. Casting `'penetration_test'` raises `invalid input value for enum "EvidenceFormType"`.

Fixed in bc952b9 — the filters now use the stored labels (`penetration-test`, `network-diagram`, `rbac-matrix`).

My original verification missed it because I tested the jsonb transform without the `formType` predicate. Re-verified properly by running the migration file itself against rows seeded back into the pre-fix shape: both convert to the declared key, the file object is preserved, and a second run is a no-op (all `UPDATE 0`).

PR #3576, opened 2026-09-22
fix(security): P1 sweep - GH-273, GH-257, GH-103, GH-101, GH-100, GH-093, GH-079, GH-078, GH-053, GH-054

dennisofficial:
Draft PR fixing 10 confirmed P1 findings from the Glass House sweep (GH-273, GH-257, GH-103, GH-101, GH-100, GH-093, GH-079 + duplicate GH-245, GH-078, GH-053, GH-054). One commit per finding. All were re-validated against `origin/main` HEAD before fixing — none were stale.

## GH-273 — mcp-server `serve`: wildcard CORS + 0.0.0.0 bind + static key fallback

**Wrong:** `serve` bound 0.0.0.0:2718, set `Access-Control-Allow-Origin: *` on every response, and POST /mcp had no auth — `buildSDK` silently falls back to the CLI `--apikey`. Any website the operator visited could drive the local MCP server with their full-privilege key; the LAN could reach it too.
**Changed:** bind 127.0.0.1 by default (new `--host` flag to opt out); wildcard CORS middleware removed (MCP HTTP clients are not browsers); POST /mcp rejects requests with a non-localhost `Origin` header (browsers always send one; header-less local MCP clients keep working with `--apikey`). `--disable-static-auth` semantics unchanged.
**Verified:** `tsc --noEmit` clean in apps/mcp-server (own install — it's outside the bun workspaces).

## GH-257 — task status route returned any run's output to any authenticated user

**Wrong:** `GET /api/tasks/[taskId]/status` called `runs.retrieve(taskId)` and returned output/status/error with no ownership check — cross-tenant read of AI-generated policy/questionnaire/vendor content given a run id.
**Changed:** every user-triggerable `tasks.trigger` call site in apps/app now tags the run with the owning organizationId; the route requires `run.tags` to include the caller's active org and returns the same 404 as a missing run otherwise. `research-vendor` (shared GlobalVendors data, reachable pre-onboarding) is tagged when an active org exists.
**Verified:** new `route.test.ts` (5 cases: unauthenticated, no org, cross-tenant 404, untagged 404, own-org 200) passes; `tsc --noEmit` clean on all 18 touched files.

## GH-103 — Upstash Vector filter injection via organizationId

**Wrong:** `findSimilarContent`/`findSimilarContentBatch` built the metadata filter by raw interpolation (`organizationId = "${organizationId}"`); a value containing `"` could append arbitrary clauses (`OR organizationId GLOB "*"`) and dump every tenant's RAG chunks. The org id is reachable from the `answer-question` task payload.
**Changed:** strict allowlist (`/^[a-zA-Z0-9_-]+$/`, the prefixed-CUID shape) with fail-closed guard before the filter string is built — in both query paths and the same-pattern sink in `sync-organization.ts` (embedding readiness check). The apps/app copy does client-side post-filtering only and is not vulnerable.
**Verified:** 14/14 jest tests pass in src/vector-store, incl. new tests replaying the literal exploit payload against both functions.

## GH-101 — Sentry Session Replay unmasked (maskAllText: false)

**Wrong:** production replay ran with `maskAllText: false` / `blockAllMedia: false` (10% of sessions, 100% of error sessions) and the `sentry-mask` escape hatch was used in zero of ~1470 components — verbatim capture of customer compliance data.
**Changed:** `maskAllText: true`, `blockAllMedia: true` in apps/app and apps/portal; `data-sentry-unmask` is now the opt-in for provably safe elements.
**Verified:** typecheck clean on both files.

## GH-100 — hardcoded 'fallback-secret' HMAC key for unsubscribe tokens

**Wrong:** `@trycompai/email` derived the unsubscribe-token HMAC key from `UNSUBSCRIBE_SECRET || AUTH_SECRET || 'fallback-secret'`; apps/api never sets either and verifies the unauthenticated `POST /v1/email/unsubscribe` with it — every token forgeable from source.
**Changed:** fallback literal removed; the secret resolves lazily and token generation throws when unconfigured (fail closed; importing the package never throws). `UNSUBSCRIBE_SECRET` documented in apps/api/.env.example. The separate `apps/app/src/lib/unsubscribe.ts` implementation was checked — it has no fallback and needs no change.
**Verified:** all call sites are request/task-time, so lazy resolution is safe; packages/email typecheck shows only pre-existing `render.test.tsx` errors.

## GH-093 — portal policy-signature endpoints write to any org's policies

**Wrong:** `mark-policy-completed` looked the member up with no org filter and the policy by bare id; `accept-policies` verified the memberId but pushed it into `signedBy` of policies in any organization.
**Changed:** `mark-policy-completed` resolves the member within the policy's organization (403 otherwise — also fixes the arbitrary-membership pick for multi-org users); `accept-policies` skips policies outside the member's organization.
**Verified:** `tsc --noEmit` clean on all touched portal files.

## GH-079 (+ duplicate GH-245) — portal policy-pdf-url trusts versionId

**Wrong:** the `versionId` branch did a bare `findUnique({ where: { id: versionId } })` and presigned whatever `pdfUrl` came back — a 15-minute presigned GET for another org's policy PDF. The main app's sibling implementation has exactly this check and comments why.
**Changed:** version lookup scoped to the already-org-validated `policyId` (`where: { id: versionId, policyId }`); mismatches fall through to the existing "No PDF found." response.
**Verified:** typecheck clean.

## GH-078 — trust questionnaire upload ignores securityQuestionnaireEnabled

**Wrong:** `POST /v1/questionnaire/parse/upload/token` validated the trust access token and ran RAG auto-answering without checking `Trust.securityQuestionnaireEnabled` — orgs that disabled the AI questionnaire were still fully served to any token holder.
**Changed:** flag checked after token validation (new `isSecurityQuestionnaireEnabledForOrganization` on TrustAccessService, defaulting to enabled when no Trust row exists, matching the public-overview helper); 403 when disabled. This is the only `@Public()` token-authenticated endpoint in the controller.
**Verified:** 92/92 jest tests pass in src/questionnaire, incl. 3 new cases (disabled → 403 and `autoAnswerAndExport` never called; enabled → proceeds; no Trust row → proceeds).

## GH-053 — organization logo is an unvalidated S3 key that gets presigned

**Wrong:** `UpdateOrganizationDto.logo` was a bare string, stored verbatim, and `GET /v1/organization` presigned it as a raw S3 key against the shared org-assets bucket (which also holds trust documents) — cross-tenant read for any org owner who knows a key. The Next.js layout duplicated the unchecked presign.
**Changed:** update rejects logo keys not prefixed with the org's own id (empty/null still clears); `getLogoSignedUrl` takes the organizationId and returns null for out-of-org keys; the app layout skips presigning out-of-org logos. Legit keys are `${organizationId}/logo/...` (written by `uploadLogo`).
**Verified:** 33/33 jest tests pass in src/organization incl. new cross-org rejection cases; app typecheck clean on the layout.

## GH-054 — evidence-form fileKey accepted arbitrary keys, presigned on read

**Wrong:** `fileKey: z.string().min(1)` was stored verbatim in submission data and `refreshFileUrls` presigned whatever was stored against the shared attachments bucket — cross-tenant read given a leaked full key.
**Changed:** submission create rejects any file field whose fileKey doesn't start with `${organizationId}/` (legit keys are `${organizationId}/attachments/evidence-forms/...`); `refreshFileUrls` takes the organizationId and skips presigning (downloadUrl: null) for out-of-org keys — defense in depth for legacy rows.
**Verified:** 32/32 jest tests pass in src/evidence-forms, incl. 4 new guard tests.

## Needs human action

- **GH-100 (required before deploy):** set `UNSUBSCRIBE_SECRET` in the apps/api production environment (and anywhere `@trycompai/email` sends mail). Without it, email sending now throws instead of silently using the public fallback. Treat any secret material previously derivable as compromised: tokens minted under `fallback-secret` were forgeable by design, so invalidating them by setting the secret is the intended effect.
- **GH-101:** replays already captured in Sentry contain unmasked customer data — decide whether to purge Sentry replay history.
- **GH-257:** runs triggered before this deploy carry no org tag and will 404 on the status route — in-flight runs at deploy time need re-triggering; no data migration possible (tags live on the run).
- **GH-053:** any org whose stored `logo` doesn't start with its own id stops rendering a logo (logoUrl → null). Worth a one-off data check (`SELECT id, logo FROM organization WHERE logo IS NOT NULL AND logo NOT LIKE id || '/%'`); affected rows can re-upload.
- **GH-054:** legacy submissions holding out-of-org fileKeys now render `downloadUrl: null` instead of a presigned URL.
- **GH-273:** `serve/impl.ts` and `serve/command.ts` carry the Speakeasy "DO NOT EDIT" header but are hand-maintained in-repo (prior feature commits edit them too) — a future Speakeasy regeneration could drop this fix; consider upstreaming the hardening into the generation config.
- **GH-273 / GH-100 both note:** apps/api `.env.example` documents `UNSUBSCRIBE_SECRET=` with no value — a real secret must be generated and added to the deployed env out of band.

## Verification summary

- apps/api: `bunx jest` — src/vector-store 14/14, src/organization 33/33, src/evidence-forms 32/32, src/questionnaire 92/92, all passing; `tsc --noEmit` reports zero errors in touched files (remaining failures are pre-existing errors in unrelated spec files, identical at HEAD).
- apps/app: `bunx vitest run` new status-route test 5/5; `tsc --noEmit` clean on all touched files.
- apps/portal: `tsc --noEmit` clean on touched files.
- apps/mcp-server: `tsc --noEmit` clean (workspace-excluded package, own `bun install`).
- packages/email: typecheck shows only pre-existing `render.test.tsx` errors (confirmed identical with the change stashed).
- Lint: eslint clean on touched files except pre-existing warnings/prettier noise also present at HEAD.

**Do not merge** — draft for review per the sweep protocol.

---
## Summary by cubic
Fixes 10 confirmed P1 security findings from the Glass House sweep across `apps/api`, `apps/app`, `apps/portal`, `apps/mcp-server`, and `packages/email`.

Closes a cross-tenant read path (GH-257) by tagging every user-triggerable task run with its owning organization and requiring a matching tag on the status route. Closes S3 key validation gaps (GH-053, GH-054) so logo keys, evidence-form fileKeys, and presigned download URLs are scoped to the caller's organization. Closes Upstash Vector filter injection via `organizationId` (GH-103) with a strict allowlist. Closes the portal policy endpoints (GH-093, GH-079) by scoping lookups to the member's organization, and honors `securityQuestionnaireEnabled` on token-authenticated uploads (GH-078). Closes local MCP server exposure (GH-273) by binding to `127.0.0.1` by default and rejecting non-localhost `Origin` headers. Removes the hardcoded `fallback-secret` HMAC key for unsubscribe tokens (GH-100) and masks all text/media in Sentry Session Replay (GH-101).

**Migration**
- Set `UNSUBSCRIBE_SECRET` in the `apps/api` production environment before deploy, otherwise email sending throws instead of silently using the public fallback. Tokens minted under the fallback secret were forgeable; resetting the secret invalidates them.
- Runs triggered before this deploy carry no org tag and 404 on the status route; in-flight runs at deploy time need re-triggering.
- Orgs with a stored `logo` not prefixed by their own id stop rendering a logo, and legacy evidence submissions holding out-of-org fileKeys render `downloadUrl: null`. Existing replays in Sentry contain unmasked customer data — decide whether to purge replay history.

<sup>Written for commit e2f27d659352f2689063f3f1a172c52cf4a64087. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/trycompai/comp/pull/3576?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dennisofficial:
@cubic-dev-ai review it

